PR #104, opened 2014-12-04
Mark run_tutorial_a executable same as other examples

keszybz:
Just for consistency so rpmlint doesn't complain.

### DIFF
--- a/pysb/examples/run_tutorial_a.py
+++ b/pysb/examples/run_tutorial_a.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python
+"""Run the first example in the tutorial and print the results."""
+
 from pysb.integrate import Solver
 from tutorial_a import model
 


keszybz:
Thanks.

> How does rpmlint determine that this is supposed to be an exectuable script?
It uses some (probably ugly) heuristics. In this case the file has code at the top level (as opposed to functions or classes), and it detects that.

alubbock:
@jmuhlich Any objection to merging this?

alubbock:
I don't have permission to push a commit to resolve the conflict here, so I can't merge it at present. Since this is quite an old PR, I'll close it, but if you'd still like to merge this change please open a new PR against `master` and I'll be happy to take a look. Thanks for your interest in PySB!

jmuhlich:
I was leaving it open as I had started to rethink this from a slightly higher level. Reopened as issue #413 (where I suggest a different fix).

PR #110, opened 2015-01-07
Add sphinx.ext.autosummary to conf.py

keszybz:
sphinx otherwise does not understand ..autosummary.

pysb/core.py:docstring of pysb.core.Observable:32: ERROR: Unknown directive type "autosummary".


jmuhlich:
Fantastic, we were trying to figure out how to resolve that error!

PR #111, opened 2015-01-07
Fix exception detail in docstring

keszybz:
Apparently " for array" was removed from the exception string
in the scipy version that I have.


jmuhlich:
Looks good, except please eliminate the extraneous whitespace that was introduced before the #doctest directives.

jmuhlich:
Nevermind, I took care of the cleanup as well as rebased against master in #116.

Issue #101, opened 2014-11-24
python 3 support


keszybz:
Are there any plans? Would potential patches to make code python2/3 compatible be accepted?

alubbock:
One workaround for the lack of `<>` would be to use named infix operators, like so: http://code.activestate.com/recipes/384122/

keszybz:
On Mon, Jun 29, 2015 at 03:04:42PM -0700, alubbock wrote:
> One workaround for the lack of `<>` would be to use named infix operators, like so: http://code.activestate.com/recipes/384122/

I think that's too much of hack.

jmuhlich:
If we wanted to go down that road, I'd rather go all in and use macropy to
pull off some *really* cool tricks. But I think macropy is too "out there"
to make it an absolute requirement. Even if we had optional support for it,
we would still need a more "conservative" approach to replacing <>. Perhaps
one of the bitwise operators, as they have the right precedence level. "&"
could work, with the mnemonic "it goes forwards AND backwards". This would
also free up the equality/comparison operators, making them usable for
actual comparisons again.

Another option is to remove the whole notion of reversibility from
RuleExpression, only provide the >> operator, and require the reversibility
to be specified as an explicit argument to the Rule constructor with a
default of True. So an irreversible rule would look like this: Rule(A>>B,
kf, reversible=False) .

keszybz:
Hm, the operator issue notwithstanding, what about merging #112 (or something similar), without the operator change for now. The result would not be Python3 compatible, but would be very close, and Python3 users could do the patch to change the operator on their own. If that would be acceptable, I'd update #112 (I presume it doesn't apply cleanly anymore) and remove operator change from it.

jmuhlich:
Yes, let's do that.

PR #112, opened 2015-01-07
Python 3 compatibility

keszybz:
This pull request allows pysb to be used under Python 3.4 (and probably 3.3, but I didn't test that). All tests pass, and things seem to generally work the same as under Python 2.7.

The code is made both Python 2 and 3 compatible. This is the modern approach to doing the conversion. A single codebase makes everything easier and faster. Since the pysb codebase is rather clean, there's no need to use helper modules like six.

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -29,7 +29,7 @@ sampled at intervals of 10 seconds::
     >>> t = [0, 10, 20, 30, 40, 50, 60]
     >>> solver = Solver(model, t)
     >>> solver.run()
-    >>> print solver.y[:, 1]
+    >>> print(solver.y[:, 1])
     [   0.   30.   60.   90.  120.  150.  180.]
 
 Creating a model
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -273,7 +273,7 @@ parameters. The rule interactions make use of the following
 operators::
 
    *+* operator to represent complexation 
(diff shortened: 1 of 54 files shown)


keszybz:
This should fix issue #101 :)

jmuhlich:
This is an amazing amount of work, thank you so much. It will take a little time to review, but we will get to it.

jmuhlich:
OK, I finally reviewed this (although I'll admit I didn't check every line in the print_function one... :) and I just have a small issue with the scipy.weave fix. I'll comment directly on that commit and we can resolve it.

As discussed at #101, do please back out the != operator change. Is there an easy mechanical way for a Python3 user to apply that fix locally?

keszybz:
>  I'll comment directly on that commit and we can resolve it.
Did you comment? I don't see the comment, fwiw.

>  Is there an easy mechanical way for a Python3 user to apply that fix locally?
Partially. Changes to code can be done with 2to3 operator fixer. That wouldn't change documentation. But having just the code changes is enough to use it.

keszybz:
I looked into changing the weave import to be in the try_block, but it makes things more complicated... scipy.weave.build_tools.CompileError is not available if the import fails, so a nested try..except block is needed, etc. I tweaked the implementation from 0441755 to be a bit more readable, but I'm leaving it mostly as is.

I'll submit a new pull request with updated code.

PR #117, opened 2015-02-26
Support for Expressions in Kappa and other minor fixes

ryants:
This series of commits enables the use of named parameters and functions in the Kappa language (these are both implemented in Kappa via the "%var" declaration).  It also fixes a syntax error that arises when generating Kappa patterns denoting a site which does not care about its bond state (i.e. the WILD bond state).  The syntax present in the current version produces the string "site~state!?" and the correct syntax is "site~state?".  Finally, this pull request allows formation of patterns in which a site contains both an explicit state and an arbitrary bond to another agent (the ANY bond state), which is not currently allowed in PySB, despite its legality in both BNGL and Kappa pattern matching; for example: "site~state!+" in BNGL and "site~state!_" in Kappa.


johnbachman:
This is great, I am taking a look at this now. To make this easier, do you think you could write some tests showing showing where the old code failed and the new code works? You could create a file test_kappa.py under pysb/tests and follow the examples given in test_bng.py (including the use of the @with_model decorator to set up and tear down models). Thanks!

ryants:
No problem!  I assume I can just add the tests to this pull request via commits to my forked master branch?

jmuhlich:
Yes, GitHub will update the PR magically when you push more commits to that
branch.

ryants:
Finally got around to writing those tests (and I also added Kappa dependencies to .travis.yml).  Let me know if you have questions or concerns!

johnbachman:
Sweet, will take a look. Thanks!

jmuhlich:
This has been integrated via #188.

PR #118, opened 2015-04-15
Fixes to _verify_sites allowing it to work on complexes

jmuhlich:


jmuhlich:
:+1: from me. Will merge after CI build finishes.

PR #120, opened 2015-06-11
Calculate Jacobian of ODEs analytically and supply to integrator

johnbachman:
This feature uses Sympy to analytically calculate the Jacobian of the ODEs. Like the RHS function, the Jacobian is implemented as a callable that returns a matrix, and like the RHS it can also be inlined using scipy.weave. When the Jacobian is not explicitly supplied to the integrator (e.g., VODE), it is approximated by finite-differences calculations. Using the analytical Jacobian, inlined, results in small but consistent speedups for simulations of the EARM and Robertson models. Hopefully it will also help minimize convergence failures during parameter estimation runs!

More comments on specific changes can be found in the commit messages.

### DIFF
--- a/pysb/integrate.py
+++ b/pysb/integrate.py
@@ -86,16 +86,28 @@ class Solver(object):
 
         pysb.bng.generate_equations(model)
 
+        def eqn_substitutions(eqns):
+            """String substitutions on the sympy C code for the ODE RHS and
+            Jacobian functions to use appropriate terms for variables and
+            parameters."""
+            # Substitute 'y[i]' for 'si'
+            eqns = re.sub(r's(\d+)',
+                          lambda m: 'y[%s]' % (int(m.group(1))), eqns)
+            # Substitute expanded parameter formulas for any named expressions
+            for e in model.expressions:
+                eqns = re.sub(r'\b(%s)\b' % e.name,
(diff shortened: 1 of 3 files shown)


pysb/integrate.py
```diff
@@ -82,20 +87,36 @@ def _test_inline():
                     distutils.errors.CompileError, ImportError):
                 pass
 
-    def __init__(self, model, tspan, integrator='vode', **integrator_options):
+    def __init__(self, model, tspan, use_jacobian=True, integrator='vode',
```
jmuhlich: should default to False (also fix in docstring)

pysb/integrate.py
```diff
@@ -145,7 +217,14 @@ def rhs(t, y, p):
         else:
             self.yexpr = numpy.ndarray((len(tspan), 0))
         self.yobs_view = self.yobs.view(float).reshape(len(self.yobs), -1)
-        self.integrator = ode(rhs).set_integrator(integrator, **options)
+        # Initialize an instance of scipy.integrate.ode
+        if self._use_jacobian:
```
jmuhlich: In conditional, just set value of local "jac" variable, then pass that in a single call to ode(). You'll need to figure out the default value of jac (possible None?) in the ode definition.

Issue #151, opened 2015-09-18
Expressions consisting of only a single parameter aren't converted into correct BNGL


johnbachman:
If you have a model that contains an expression that is mapped just to a specific parameter, as in the following model:

    from pysb.bng import generate_network
    from pysb import *

    Model()
    Monomer('A')
    Parameter('k1', 1.0)
    Expression('expr', k1)
    Initial(A(), expr)
    Rule('Adeg', A() >> None, Parameter('kdeg', 1.0))

    generate_network(model)

You get the following error from BNG:

    File "/Users/johnbachman/Dropbox/1johndata/programs/pysb/pysb/bng.py", line 258, in
    generate_network
        raise GenerateNetworkError(p_out.rstrip()+"\n"+p_err.rstrip())
    pysb.bng.GenerateNetworkError: BioNetGen version 2.2.5
    Reading from file test_expressions_71172_876_temp.bngl (level 0)
    ABORT: Expecting operator argument in expr=Parameter('k1', 1.0) at 'k1', 1.0) at line 5

The BNGL contains the following Parameters block:

    begin parameters
      k1           1.000000e+00
      kdeg         1.000000e+00
      __source_0   1.000000e+00
      expr         Parameter('k1', 1.0)
    end parameters

So clearly, the Expression 'expr' isn't getting generated properly.

As for why it's useful to have an expression that is mapped to a single parameter, there are cases where in some models you want to have a rate scaled in a certain way, and in other models not. When using model or builder subclassing strategy, you can have each model reference the same expression and have the subclasses scale the parameters appropriately.

PR #152, opened 2015-09-18
Change stringification of Expressions to allow expressions with one Parameter (or Observable)

johnbachman:
This fixes issue #151. The problem was that sympy_to_muparser in pysb.generator.bng was calling the str() function on Expression.expr directly, which works when the Expression is some kind of algebraic formula, but not when it's a single Parameter or Observable (because the Parameter/Observable's own repr method supersedes sympy.Symbol). To fix this I added a function that is explicitly for getting the string representation of the Expression formula.

Added two new tests to test_bng.py, which pass as a result of this patch.


johnbachman:
Actually I've just discovered some more bugs, don't anybody merge this just yet.

PR #153, opened 2015-09-21
Fix expressions

johnbachman:
This fixes issues #147 and #151. There were two problems: the first was that sympy_to_muparser in pysb.generator.bng was calling the str() function on Expression.expr directly, which works when the Expression is some kind of algebraic formula, but not when it's a single Parameter or Observable (because the Parameter/Observable's own repr method supersedes sympy.Symbol). To fix this I added a function that is explicitly for getting the string representation of the Expression formula.

The other problem was that the call to expand_expr(self.model).evalf(subs=subs) in integrate.py had a syntax error--the argument self.model was erroneous (issue #147). Also this call was using the wrong substitution dict--the substitution dict was keyed by parameter name, rather than by symbol (Parameter object).

The new file test_expressions.py checks that Expressions of 1 or 2 Parameters or Observables can both be generated by BNG and integrated correctly by integrate.py.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -786,9 +786,9 @@ class Parameter(Component, sympy.Symbol):
 
     def __repr__(self):
         return  '%s(%s, %s)' % (self.__class__.__name__, repr(self.name), repr(self.value))
-    
+
     def __str__(self):
-        return  '%s(%s, %s)' % (self.__class__.__name__, repr(self.name), repr(self.value))
+        return  repr(self)
 
 
 
@@ -842,7 +842,6 @@ class Compartment(Component):
     def __repr__(self):
         return  '%s(name=%s, parent=%s, dimension=%s, size=%s)' % \
             (self.__class__.__name__, repr(self.name), repr(self.parent), repr(self.dimension), repr(self.size))
-
 
 
 class Rule(Component):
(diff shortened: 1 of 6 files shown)


johnbachman:
OK this looks good now. If there are no objections I'll merge it in a day or two.

alubbock:
Looks good. Only thing I can think of would be to add some expected failures to the expressions unit tests. The only (trivial) one that springs to mind is an undefined symbol, but maybe you guys can think of more scenarios?
```
@raises(NameError)
def test_expression_with_undefined_symbol():
    Expression('e1', o1)
```

alubbock:
@lh64 mentioned that it'd be a good idea to test nested expressions too. Here's an example you could use.
```python
@with_model
def test_nested_expression():
    Monomer('A')
    Monomer('B')
    Parameter('k1', 1)
    Observable('o1', B())
    Expression('e1', o1*10)
    Expression('e2', e1+5)
    Initial(A(), Parameter('A_0', 1000))
    Initial(B(), Parameter('B_0', 1))
    Rule('A_deg', A() >> None, e2)
    generate_equations(model)
    t = np.linspace(0, 1000, 100)
    sol = Solver(model, t, use_analytic_jacobian=True)
    sol.run()
```

johnbachman:
@alubbock @lh64 I like the nested expression test. I put your version in verbatim and it passes. I left out the NameError one because it is not a test of expressions per se (the error will simply trigger because o1 is an undeclared variable).

jmuhlich:
reviewed :+1:

PR #159, opened 2015-09-25
Unit tests for pysb/util.py (only exported functions)

### LINKED ISSUE #158: Create test_util.py
We need a unit test suite for `util.py`. Currently, it has only ~26% code coverage.

alubbock:
Closes #158. Only the two exported functions -- `alias_model_components` and `rules_using_parameter` -- are tested for now.


pysb/tests/test_util.py
```diff
@@ -0,0 +1,41 @@
+from pysb.core import *
```
jmuhlich: I think we should move to eliminate "import *" from our own code eventually but I've written enough of those in this codebase that I'm not going to fault you for this one. :) Pass!

pysb/tests/test_util.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_alias_model_components():
+    """
+    Tests that alias_model_components() exports to the global namespace
+    """
+    m = Model(_export=False)
+    m.add_component(Monomer('temp_monomer_A', _export=False))
+    try:
+        temp_monomer_A
+        raise NameError('Monomer A should not be defined in global namespace')
```
jmuhlich: This won't fail if temp_monomer_A *is* defined! You should use nose's `assert_raises` to ensure that expected exceptions are raised. It would look something like this:

```
from nose.tools import assert_raises
assert_raises(NameError, lambda: temp_monomer_A)
```

The trick is that the second arg to assert_raises is a callable. I used a lambda to create a callable that just evaluates the global in question, which should raise NameError.

Also, why not just call the monomer 'A'?

jmuhlich: Actually I just realized an even simpler solution is to avoid raising NameError at all and just check `globals()`:

```
assert 'temp_monomer_A' not in globals()
```

pysb/tests/test_util.py
```diff
@@ -0,0 +1,41 @@
...
+    except NameError:
+        # We expect this exception due to _export=False
+        pass
+    alias_model_components(m)
+
+    # Use global version of temp_monomer_A (needed to delete it)
```
jmuhlich: You can just delete straight out of globals with `del globals()['temp_monomer_A']` so this `global` declaration isn't needed.

pysb/tests/test_util.py
```diff
@@ -0,0 +1,41 @@
...
+    # Delete the monomer to cleanup the global namespace
+    del temp_monomer_A
+
+
+def test_rules_using_parameter():
+    """
+    Tests for rules_using_parameter() in pysb.util
+    """
+    # Get rules matching parameter name
+    r = rules_using_parameter(earm_1_0.model, 'kc1')
```
jmuhlich: This probably shouldn't depend on earm_1_0 but rather define its own little model to test. I wrote a decorator `pysb.testing.with_model` which lets you just start declaring components in a test function without declaring the model or worrying about cleaning up at the end, so writing mini-models in test functions is not so painful. See the tests in test_core for many examples.

The variable name `r` is not great, as the return value is a ComponentSet not a Rule. `rules` or `components` would be better.

pysb/tests/test_util.py
```diff
@@ -0,0 +1,41 @@
...
+    assert isinstance(r.get('produce_DISC'), Rule)
```
jmuhlich: The idiomatic way to get an element from a ComponentSet by name is to index it like a dict: `r['produce_DISC']`. Is there another reason to use `.get` here instead?

pysb/tests/test_util.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # Get rules by supplying Parameter object directly
+    r = rules_using_parameter(earm_1_0.model,
+                              earm_1_0.model.parameters.get('kc1'))
+    assert isinstance(r.get('produce_DISC'), Rule)
```
jmuhlich: Your editor seems to be stripping the newline at the end of files. I've seen this on a bunch of other files in your PRs, both new files and existing ones that were edited. Can you check on that and disable it, and fix the files touched in this PR?

pysb/util.py
```diff
@@ -86,4 +94,4 @@ def load_params(fname):
     temparr = numpy.loadtxt(fname, dtype=([('a','S50'),('b','f8')]), delimiter=',') 
     for i in temparr:
         parmsff[i[0]] = i[1]
-    return parmsff
+    return parmsff
```
jmuhlich: Fix stripped newline.

jmuhlich:
Thanks Alex.

PR #162, opened 2015-09-26
Correct Annotation tags' argument order (subject, object, predicate)

alubbock:


jmuhlich:
Oh dear. Good catch. We made the Annotation API take that odd order (as opposed to **subject, predicate, object** which would translate more naturally from English) so that the expected common case of "is" could be implemented by a default value on the predicate arg. I guess we could have made that pattern work with the predicate in the middle, but the code would have been trickier and the \__init__ signature less informative:
```
def __init__(self, subject, arg2, arg3=None):
    ...
    if arg3 is None:
        self.predicate = 'is'
        self.subject = arg2
    else:
        self.predicate = arg2
        self.subject = arg3
    ...
```
If we all like that more, we can always change it for v2 when we can break APIs. :)

PR #167, opened 2015-09-29
Add python3 support to Travis config

### LINKED ISSUE #166: No __ version __ constant in root pysb module
It is quite common for Python modules to have __ version __ constant defined in the root module namespace to enable version checking and tracking.
It seems to be lacking from PySB.

jmuhlich:
This lets us test Python 2.7 and 3.4 in parallel on Travis. It also switches our setup.py from distutils to setup tools, which it turns out is not strictly required but still nice, and brings in [Versioneer](https://pypi.python.org/pypi/versioneer/) to make our version strings nicer and PEP440 compliant to boot (my old solution had problems on Travis for some reason). Also fixes #166 .


alubbock:
I've added a couple of specific comments on the commits, and I didn't extensively check the ez_setup.py or versioneer code (assume it's boilerplate?). Otherwise, looks good!

jmuhlich:
Oops, I did a rebase and push -f without realizing it would erase your commit comments. Sorry! I agree with all of your comments and will re-add some notes reminding me what I agreed to do for them.

jmuhlich:
@alubbock OK, take a look again.

alubbock:
Looks good to merge @jmuhlich. Thanks!

PR #168, opened 2015-10-02
New BioNetGen interfaces for console and file based access

alubbock:
I've added two new classes, `BngFileInterface` and `BngConsole`, for interacting with BioNetGen. Both extend an abstract base class `BngBaseInterface`. I've adjusted `generate_network` and `run_ssa` to use `BngFileInterface`.

Both `BngFileInterface` and `BngConsole` have very similar API and usage. The difference is that `BngConsole` executes commands instantly and waits for the response using the `pexpect` library and the BNG Console interface. `BngFileInterface` buffers all BNG commands and outputs them to the .bngl file for execution in one go. `BngFileInterface` is typically faster (no repeated command/response lag) and cross-platform; `BngConsole` is great for interactive use and development, but the `pexpect` library doesn't support Windows.


johnbachman:
Looks good to me. @jmuhlich, do you want to take a look?

.travis.yml
```diff
@@ -12,7 +12,7 @@ before_install:
 - conda install --yes python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib sympy networkx
   nose h5py
 - pip install -i https://pypi.binstar.org/pypi/simple pygraphviz==1.3rc1
-- pip install python-coveralls
+- pip install python-coveralls pexpect
```
jmuhlich: Also pedantic: I think anaconda offers pexpect (see http://docs.continuum.io/anaconda/pkg-docs), so ideally we'd install that with conda rather than pip.

pysb/bng.py
```diff
@@ -91,14 +92,260 @@ class GenerateNetworkError(RuntimeError):
...
+    def __init__(self, model, verbose=False, cleanup=False,
+                 output_prefix=None):
+        self.verbose = verbose
+        self.cleanup = cleanup
+        self.output_prefix = 'tmpBNG' if output_prefix is None else \
+            output_prefix
+        self.generator = BngGenerator(model)
+        self._check_model()
+
+    def __enter__(self):
```
jmuhlich: Shouldn't this also be annotated with `abc.abstractmethod`?

pysb/bng.py
```diff
@@ -91,14 +92,260 @@ class GenerateNetworkError(RuntimeError):
...
+            An argument to a BNG action call
+        """
+        if isinstance(param, str):
+            return '"%s"' % param
+        if isinstance(param, bool):
+            return 1 if param else 0
+        return param
+
+    @abc.abstractmethod
+    def action(self, action, **kwargs):
```
jmuhlich: This is a misuse of `abstractmethod` -- this `action` and its implementation in subclasses do different things. This one should have a different name, like `format_action`, and be a `classmethod`. Then `action` would be an empty `abstractmethod`. The subclasses would then call `self.format_action` to build the `action_args` instead of `super(...).action`.

pysb/bng.py
```diff
@@ -91,14 +92,260 @@ class GenerateNetworkError(RuntimeError):
...
+            # Generate BNGL file
+            with open(self.bng_filename, 'w') as bng_file:
+                bng_file.write(self.generator.get_content())
+                bng_file.write(self.command_queue.getvalue())
+
+            p = subprocess.Popen(['perl', _get_bng_path(), self.bng_filename],
+                                 cwd=os.path.dirname(self.bng_filename),
+                                 stdout=subprocess.PIPE,
+                                 stderr=subprocess.PIPE)
+            if self.verbose:
```
jmuhlich: Will `verbose=True` eat the output from BNG such that a potential `GenerateNetworkError` below won't have any content to work with? I haven't pulled this code down to test that yet.

jmuhlich: I realize this is effectively the same code that was in `generate_network` before but I missed this issue when we reviewed the big dev branch merge.

alubbock:
I think I've addressed all your feedback @jmuhlich - let me know if there's anything I missed. The `verbose=True` argument in `BngFileInterface` doesn't stop the output from being captured if there's an error, but it did make me realize `BngConsole` would hide any issues unless `verbose=True` was set. I've added new warning output which passes on any BNG warnings encountered in 0e2262e.

jmuhlich:
Hmm, it's good to note that your "allow synthesis instead of initial conditions" fix is rolled up in here too.

PR #169, opened 2015-10-09
fixed a bug in the creation of bidirectional reactions.

ortega2247:
this resolves a problem we were having with render_reactions.

### DIFF
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -419,26 +419,25 @@ def _parse_reaction(model, line):
     model.reactions.append(reaction)
     # bidirectional reactions
     key = (reactants, products)
-    reaction_bd = reaction_cache.get(key)
-    if reaction_bd:
+    key_reverse = (products, reactants)
+    if key in reaction_cache:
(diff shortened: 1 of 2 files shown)


pysb/bng.py
```diff
@@ -331,7 +331,7 @@ def _parse_netfile(model, lines):
             _parse_reaction(model, line)
         # fix up reactions whose reverse version we saw first
         for r in model.reactions_bidirectional:
-            if r['reverse']:
+            if all(r['reverse']):
```
alubbock: This is needed because `r['reverse']` is now a tuple. Whether to check `all()` or only the first entry in `r['reverse']` is a matter of preference of how to canonicalize which way round reversible rules are defined. Any opinions on this?

jmuhlich:
I think the first rule in the model should "win" in terms of deciding which way is "forward" in a reaction in `reactions_bidirectional`. It looks like this code does implement that.

jmuhlich:
Thanks Oscar and Alex! :+1:

PR #171, opened 2015-10-16
Allow models with no initial conditions if synthesis rule(s) are present

alubbock:
BioNetGen allows models without initial conditions, provided a zero-order synthesis rule is specified. This PR implements that feature in PySB.


alubbock:
I've implemented this in commit 2c82cd5 in PR #168 instead, as the relevant code is largely refactored anyway. Closing.

PR #180, opened 2015-11-16
Improved BNG temp file handling & support for sequential BNG simulations

### LINKED ISSUE #179: Use mkdtemp rather than NamedTemporaryFile in BNG interface code
To encapsulate the temporary input/runtime/output files needed for interfacing with and running BNG, doing everything in a directory created with `tempfile.mkdtemp` would be more secure than a series of files based on just a `NamedTemporaryFile` base name. (Thanks to @dkoes for discussion in #178 )

alubbock:
This PR improves BioNetGen temp file handling by securely creating a temporary directory for each interface instance to BNG with python's `tempfile.mkdtemp` (fixes #179). This required adding in an explicit `execute()` step to `BngFileInterface`, in order to allow the user to read in files within the interface's `with(...)` context before it exits, when cleanup (deleting the temp directory) is done.

As a side effect, this enables sequential sets of actions to be run in multiple `execute()` steps. Since network generation can be costly for large networks, I also added an `execute(reload_netfile=True)` option, which is useful for e.g. sequential simulations in BNG using the same reaction network (see the new unit test in pysb/tests/test_bng.py).

### DIFF
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -11,6 +11,7 @@ import tempfile
 from pkg_resources import parse_version
 import abc
 from warnings import warn
+import shutil
 
 try:
     from cStringIO import StringIO
@@ -120,9 +121,7 @@ class BngBaseInterface(object):
         return
 
     def _delete_tmpdir(self):
-        for f in os.listdir(self.base_directory):
-            os.unlink(os.path.join(self.base_directory, f))
-        os.rmdir(self.base_directory)
+        shutil.rmtree(self.base_directory)
 
     def _check_model(self):
         """
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -109,9 +109,8 @@ class BngBaseInterface(object):
         self.generator = BngGenerator(model)
         self._check_model()
 
-        self.base_directory = tempfile.mkdtemp(prefix=output_prefix,
+        self.base_directory = tempfile.mkdtemp(prefix=self.output_prefix,
                                                dir=output_dir)
-        self.bng_filename = self.base_filename + '.bngl'
 
     def __enter__(self):
         return self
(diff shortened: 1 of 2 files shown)


jmuhlich:
@alubbock I think I fixed the travis/coveralls issue. Could you retrigger the travis build so it updates here?

pysb/bng.py
```diff
@@ -116,6 +119,11 @@ def __enter__(self):
     def __exit__(self):
         return
 
+    def _delete_tmpdir(self):
+        for f in os.listdir(self.base_directory):
+            os.unlink(os.path.join(self.base_directory, f))
+        os.rmdir(self.base_directory)
+
```
jmuhlich: I suggest replacing these 3 lines with [shutil.rmtree](https://docs.python.org/2/library/shutil.html#shutil.rmtree).

alubbock: Thought about it - it seemed safer to do it this way rather than `shutil.rmtree` as we don't want to delete recursively (with the current setup, anyway). I guess that may change and `rmtree` is cleaner though. I'll make the change.

jmuhlich: I can't see why we wouldn't be clear to blow a temp working dir away. If the user wants to keep the temp output files around, they need to pass `cleanup=False`.

jmuhlich: If you want to explicitly call unlink on a list of known BNG-produced filenames, then rmdir the parent, I could get behind that too. But that middle ground of reimplementing a one-level-only rmtree seems like the worst of both worlds (*won't* error if other unexpected files somehow end up in the temp dir, and *will* error if a later version of BNG starts doing stuff in subdirectories).

alubbock: The explicit delete known files approach is appealing, but I tend to agree that deleting a recently created temp directory not accessible by other users should be safe, and is lower maintenance. Will go with `rmtree`.

PR #182, opened 2015-11-30
Fix species name fixup in matlab exporter

### LINKED ISSUE #170: matlab exporter produces invalid variable names in odes
There was a change at some point in the PySB internal ODE generation such that the species symbols s0, s1, etc. now have two underscores prepended (to prevent collisions with parameter names etc. I think), but the matlab exporter was not updated to strip those. So it converts __s0 to __y(1) and so on, yet leading underscores are not allowed in matlab symbols so the resulting .m file is invalid. There is a quick fix, which is to make the regex that matches `s(\d+)` also match the leading underscores, thus stripping them. I could submit a trivial PR right now, but I thought this would be a good time to figure out how to write a test for the matlab exporter to catch these things in the future. I suspect it won't fly on Travis (unless we can make it work with Octave too!) but we could at least run such test(s) opportunistically if matlab is available.

jmuhlich:
This wasn't updated when our internal species symbols gained the "__"
prefix, so the prefixes were left out of the regex that matches species
terms. The result was terms like "__y(0)" in the resulting matlab code
which is both incorrect (should be "y(0)") and also invalid matlab syntax
(no leading underscores on symbol names).

Closes #170.

### DIFF
--- a/pysb/export/matlab.py
+++ b/pysb/export/matlab.py
@@ -256,9 +256,9 @@ class MatlabExporter(Exporter):
         # Flatten to a string and add correct indentation
         odes_str = ('\n'+' '*12).join(odes_species_list)
 
-        # Change species names from, e.g., 's(0)' to 'y0(1)' (note change
+        # Change species names from, e.g., '__s(0)' to 'y0(1)' (note change
         # from zero-based indexing to 1-based indexing)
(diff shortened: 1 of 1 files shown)


jmuhlich:
Credit to @subkar for the fix!

PR #188, opened 2015-12-20
Kappa updates (ANY/WILD, Expressions, KaSa, tests)

johnbachman:
This (long overdue) pull request is based off of @ryants branch (pull request #117) which added additional Kappa functionality, including ANY and WILD bonds and Expressions, and some new tests.

In addition to these changes, I updated kappa.py to use KaSa for all static analysis rather than complx, which is now deprecated. I also fixed what was previously the "sympy_to_muparser" function in the generator/kappa.py file to make it match our current approach in generator/bng.py.

One thing to note is that Kappa 4.0, which is currently under development, will use a different syntax for observables. I updated the necessary code in 828492259e14f but then commented it out in 6204902 because this new syntax causes problems for some versions of KaSim3. When KaSim 4.0 is stable, we can uncomment these lines to make the switch to the new syntax.

I set up a Travis build procedure for testing pysb Kappa functionality, which required installing ocaml and the opam package manager for ocaml (which is then used to install a KaSim dependency, ocaml-findlib). After some fiddling around, I finally got the correct recipe for the ocaml installation/KaSim build from the .travis.yml file of the KaSim repo itself. Note that it appears that some of the source downloads from the KaSim git repo (e.g., KaSim3.5), when built, only produce a KaSim binary and not a KaSa binary, causing the tests of the KaSa functionality to fail. As a result I git cloned the KaSim source in the travis script and then did a checkout of the KaSim commit tagged "last-KaSim3" under the assumption that this will be their last KaSim 3 release before they make the (possibly breaking) changes involved in KaSim 4.

The most notable wart that remains is that the PySB Kappa tools require KaSim and KaSa to be on the user's PATH. There is currently no implementation of using an environment variable, e.g., KAPPA_HOME, to find the location of KaSim/KaSa, nor a standard set of installation locations that are searched.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -321,8 +321,9 @@ class MonomerPattern(object):
     * *list of int* : multi-bond (not valid in Kappa)
     * ``ANY`` : \"any\" bond (bound to something, but don't care what)
     * ``WILD`` : \"wildcard\" bond (bound or not bound)
-    * *tuple of (str, int)* : state with bond
+    * *tuple of (str, int)* : state with specified bond
     * *tuple of (str, WILD)* : state with wildcard bond
+    * *tuple of (str, ANY)* : state with any bond
 
     If a site is not listed in site_conditions then the pattern will match any
     state for that site, i.e. \"don't write, don't care\".
@@ -348,7 +349,7 @@ class MonomerPattern(object):
                 continue
             elif type(state) == str:
                 continue
-            elif type(state) == tuple and type(state[0]) == str and (type(state[1]) == int or state[1] == WILD):
+            elif type(state) == tuple and type(state[0]) == str and (type(state[1]) == int or state[1] is WILD or state[1] is ANY):
                 continue
             elif state is ANY:
                 continue
--- a/pysb/generator/kappa.py
+++ b/pysb/generator/kappa.py
@@ -1,4 +1,5 @@
 import pysb
+from pysb.generator.bng import sympy_to_muparser
 
 class KappaGenerator(object):
 
@@ -15,24 +16,26 @@ class KappaGenerator(object):
 
     def generate_content(self):
         self.__content = ''
-        #self.generate_parameters()
         #self.generate_compartments()
 
         # Agent declarations appear to be required in kasim
         # but prohibited in complx
         if (self.dialect == 'kasim'):
             self.generate_molecule_types() 
+            self.generate_parameters() # parameters/variables/expressions allowed in kasim
 
         self.generate_reaction_rules()
         self.generate_observables()
         self.generate_species()
 
-    #def generate_parameters(self):
-    #    self.__content += "begin parameters\n"
-    #    max_length = max(len(p.name) for p in self.model.parameters)
-    #    for p in self.model.parameters:
-    #        self.__content += ("  %-" + str(max_length) + "s   %e\n") % (p.name, p.value)
-    #    self.__content += "end parameters\n\n"
+    def generate_parameters(self):
+        #self.__content += "begin parameters\n"
+        #max_length = max(len(p.name) for p in self.model.parameters)
+        for p in self.model.parameters:
+            self.__content += ("%%var \'%s\' %e\n") % (p.name, p.value)
+        for e in self.model.expressions:
+            self.__content += ("%%var \'%s\' %s\n") % (e.name, e.expr)
+        #self.__content += "end parameters\n\n"
 
     #def generate_compartments(self):
     #    self.__content += "begin compartments\n"
@@ -53,20 +56,28 @@ class KappaGenerator(object):
     def generate_reaction_rules(self):
         #self.__content += "begin reaction rules\n"
         #max_length = max(len(r.name) for r in self.model.rules) + 1  # +1 for the colon
-        max_length = 0
         for r in self.model.rules:
             label = '\'' + r.name + '\''
             reactants_code = format_reactionpattern(r.reactant_pattern)
             products_code  = format_reactionpattern(r.product_pattern)
             arrow = '->'
+            
+            if isinstance(r.rate_forward,pysb.core.Expression) and self.dialect == 'complx':
+              raise InvalidExpressionException
+            f_rate_code = ("\'" + r.rate_forward.name + "\'" if self.dialect == 'kasim' 
+                else float(r.rate_forward.value))
             self.__content += ("%s %s %s %s @ %s") % \
-                (label, reactants_code, arrow, products_code, float(r.rate_forward.value))
+                (label, reactants_code, arrow, products_code, f_rate_code)
             self.__content += "\n"
 
             if r.is_reversible:
+              if isinstance(r.rate_reverse,pysb.core.Expression) and self.dialect == 'complx':
+                  raise InvalidExpressionException
+              r_rate_code = ("\'" + r.rate_reverse.name + "\'" if self.dialect == 'kasim' 
+                  else float(r.rate_reverse.value))
               label = '\'' + r.name + '_rev' + '\''
               self.__content += ("%s %s %s %s @ %s") % \
-                (label, products_code, arrow, reactants_code, float(r.rate_reverse.value))
+                (label, products_code, arrow, reactants_code, r_rate_code)
               self.__content += "\n"
 
         self.__content += "\n"
@@ -87,7 +98,6 @@ class KappaGenerator(object):
             raise Exception("BNG generator requires initial conditions.")
         species_codes = [format_complexpattern(cp) for cp, param in self.model.initial_conditions]
         #max_length = max(len(code) for code in species_codes)
-        max_length = 0
         #self.__content += "begin species\n"
         for i, code in enumerate(species_codes):
             param = self.model.initial_conditions[i][1]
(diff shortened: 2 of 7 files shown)


pysb/kappa.py
```diff
@@ -146,6 +122,9 @@ def contact_map(model, output_dir='.', base_filename=None, do_open=False,
 def run_complx(gen, kappa_filename, args):
     """Generalized method for passing arguments to the complx executable.
 
+    *DEPRECATED* because complx itself is deprecated. Switching over to using
+    KaSa for static analysis.
+
```
johnbachman: Perhaps I should have put a warnings.warn("DEPRECATED") warning in here?

jmuhlich: Yes, something along these lines:
```python
warnings.warn("Please use run_kasa instead", DeprecationWarning, stacklevel=2)
```

pysb/core.py
```diff
@@ -321,8 +321,9 @@ class MonomerPattern(object):
     * *list of int* : multi-bond (not valid in Kappa)
     * ``ANY`` : \"any\" bond (bound to something, but don't care what)
     * ``WILD`` : \"wildcard\" bond (bound or not bound)
-    * *tuple of (str, int)* : state with bond
+    * *tuple of (str, int)* : state with specified bond
     * *tuple of (str, WILD)* : state with wildcard bond
+    * *tuple of (str, ANY)* : state with any bond
```
jmuhlich: Huh, can't believe I forgot that one and nobody noticed for so long.

pysb/generator/kappa.py
```diff
@@ -15,24 +18,31 @@ def get_content(self):
...
-    #    self.__content += "begin parameters\n"
-    #    max_length = max(len(p.name) for p in self.model.parameters)
-    #    for p in self.model.parameters:
-    #        self.__content += ("  %-" + str(max_length) + "s   %e\n") % (p.name, p.value)
-    #    self.__content += "end parameters\n\n"
+    def generate_parameters(self):
+        names = [x.name for x in self.model.parameters] + \
+                [x.name for x in self.model.expressions]
+        for p in self.model.parameters:
+            self.__content += ("%%var: \'%s\' %e\n") % (p.name, p.value)
```
jmuhlich: Single quotes don't need to be escaped within a double-quoted string. There is a lot of that in this PR... Could probably grep for all the instances and fix them up pretty quickly.

pysb/generator/kappa.py
```diff
@@ -41,8 +51,9 @@ def generate_content(self):
     #            parent_name = ''
     #        else:
     #            parent_name = c.parent.name
-    #        self.__content += ("  %s  %d  %f  %s\n") % (c.name, c.dimension, c.size, parent_name)
-    #    self.__content += "end compartments\n\n"        
+    #        self.__content += ("  %s  %d  %f  %s\n") % \
```
jmuhlich: Unnecessary parens around the string here (although I see they were there before). Also pep8 says prefer parens over backslashes for line continuations. Could fix both issues at once like so:
```python
  self.__content += ("  %s ... " %
                     (c.name, ...))
```

pysb/generator/kappa.py
```diff
@@ -51,56 +62,88 @@ def generate_molecule_types(self):
...
         for i, code in enumerate(species_codes):
             param = self.model.initial_conditions[i][1]
-            #self.__content += ("%%init:  %-" + str(max_length) + "s   %s\n") % (code, param.name)
+            #self.__content += ("%%init:  %-" + str(max_length) + \
+            #                  "s   %s\n") % (code, param.name)
             if (self.dialect == 'kasim'):
-                # Switched from %g (float) to %d (int) because kappa didn't like scientific notation
-                # for large integers
-                self.__content += ("%%init: %d %s\n") % (float(param.value), code)
+                # Switched from %g (float) to %d (int) because kappa didn't
```
jmuhlich: This comment is out of date and misleading. Delete.

pysb/generator/kappa.py
```diff
@@ -51,56 +62,88 @@ def generate_molecule_types(self):
...
             else:
-                # Switched from %g (float) to %d (int) because kappa didn't like scientific notation
-                # for large integers
-                self.__content += ("%%init: %10d * %s\n") % (float(param.value), code)
+                if isinstance(param,pysb.core.Expression):
+                    raise KappaException("complx does not support Expressions.")
+                # Switched from %g (float) to %d (int) because kappa didn't
+                # like scientific notation for large integers
+                self.__content += ("%%init: %10d * %s\n") % \
+                                  (float(param.value), code)
```
jmuhlich: Why is param.value being converted to an float before being formatted as an integer? Can we just pass param.value as it is?

pysb/kappa.py
```diff
@@ -289,6 +268,101 @@ def run_kasim(model, time=10000, points=200, output_dir='.', cleanup=False,
...
+        * output_dict['im'] gives the influence map filename, or None if not
+          produced
+        * output_dict['cm'] gives the contact map filename, or None if not
+          produced
+    """
+
+    gen = KappaGenerator(model)
+
+    if not base_filename:
+        base_filename = '%s/%s_%d_%d_temp' % (output_dir,
```
jmuhlich: Should probably use `tempfile.mkdtemp` to create a directory without any races, even though the approach here is probably sufficient. See https://github.com/pysb/pysb/pull/180/files#diff-8f7a45525fb86b609c6d0eb1db710faeR104 for how Alex did it in the BNG interface. Note though that the default for BNG is cleanup=True because we read everything most users will want from the cdat/gdat files into memory. Also, why does `output_dir` have to be relative? Some environments (e.g. CGI scripts) may have a problem with that -- they'd want to write somewhere under `/tmp`.

pysb/kappa.py
```diff
@@ -289,6 +268,101 @@ def run_kasim(model, time=10000, points=200, output_dir='.', cleanup=False,
...
+        if p.returncode:
+            raise Exception(p.stdout.read())
+
+    except Exception as e:
+        raise Exception("Problem running KaSa: " + str(e))
+
+    finally:
+        if cleanup:
+            for filename in [kappa_filename, im_filename, cm_filename]:
+                if os.access(filename, os.F_OK):
```
jmuhlich: Better to just call `unlink` inside a try block, and `pass` on `OSError`. As written this is a race, although we'd probably never see one in practice here.

pysb/kappa.py
```diff
@@ -289,6 +268,101 @@ def run_kasim(model, time=10000, points=200, output_dir='.', cleanup=False,
...
+                    os.unlink(filename)
+
+    output_dict = {'im':im_filename, 'cm':cm_filename}
```
jmuhlich: Consider just leaving out `im` or `cm` entirely if the given output wasn't requested. Seems more pythonic to me.

pysb/tests/test_core.py
```diff
@@ -74,6 +74,11 @@ def test_monomer_as_complex_pattern():
     as_complex_pattern(A)
 
 @with_model
+def test_monomerpattern():
+    A = Monomer('A',sites=['a'],site_states={'a':['u','p']},_export=False)
+    Aw = A({'a':('u',ANY)})
```
jmuhlich: Why not `A(a=('u',ANY))` ?

pysb/tests/test_kappa.py
```diff
@@ -0,0 +1,86 @@
...
+    Monomer('A',['site'])
+    Monomer('B',['site'])
+    Initial(A(site=None), Parameter('A_0', 100))
+    Initial(B(site=None), Parameter('B_0', 100))
+    Initial(A(site=1) % B(site=1), Parameter('AB_0', 1000))
+    Rule('deg_A', A(site=pysb.WILD) >> None, Parameter('k', 1))
+    ok_(run_kasim(model, time=0, cleanup=True))
+
+def test_exported_kappa_file():
+    m = """from pysb import *
```
jmuhlich: Should just run the model definition code directly and call `pysb.export.export(model, 'kappa')`, then write *that* to `test.ka`.

pysb/tests/test_kappa.py
```diff
@@ -0,0 +1,86 @@
...
+Rule('dimerize_f',A(site='u')+A(site='u')>>A(site=('u',1))%A(site=('u',1)), kf)
+Rule('dimerize_r',A(site=('u',1))%A(site=('u',1))>>A(site='u')+A(site='u'), kr)
+Rule('degrade_dimer',A(site=('u',ANY)) >> None,kr)
+Observable('dimer',A(site=('u',ANY)))
+Observable('total_A_patterns',A(site=('u',WILD)))"""
+    wf = open('test.py','w')
+    wf.write(m)
+    wf.close()
+    commands =\
+           'python -m pysb.export test.py kappa > test.ka; KaSim -i test.ka -e 0'
```
jmuhlich: Is this not something that can be handled via `run_kasim`?

pysb/tests/test_kappa.py
```diff
@@ -0,0 +1,86 @@
...
+    # tests kappa WILD pattern and kappa syntax generation
+    ok_(subprocess.check_call(commands, shell=True) == 0)
+    subprocess.call('rm test.ka; rm test.py*', shell=True)
```
jmuhlich: Ideally these files wouldn't be generated at all (see previous comments) but if they are, they should be removed with `os.unlink`!

pysb/tests/test_core.py
jmuhlich: Ah, it's just testing the kwargs-like call mechanism. :+1:

jmuhlich:
Thanks John and @ryants, this is great!

PR #190, opened 2016-02-16
Updated simulator

JamesPino:
Two things
1. Corrected small bug in integrate class. If someone used lsoda but not analytic_jacob defaults were not properly set.

2. Updated simulator class to shrink overhead for large number of simulations. Basically just moved function calls out of long loops.


JamesPino:
Requesting merge with Lolab-VU/pysb branch.

PR #192, opened 2016-03-23
Lsoda bug fix and simulator enhancement

JamesPino:
Corrected an lsoda error that occured if _use_analytic_jacobian was False. Before it would try to assign self.jac_fn =  lambda t, y, p: jacobian(y, t, p). But jacobian is only defined if _use_analytic_jacobian =True.

The rest are enhancments to the simulator class. A lot of calls could be done once outside of the main loop, which sped up the _calc_yobs_yexp quite a bit.


jmuhlich:
Please rebase against master. This has dozens of duplicated commits and extraneous merges due to its merge history, which a rebase should clean up.

jmuhlich:
Going to rebase it and open a new PR myself. Will copy your comment there.

PR #195, opened 2016-03-30
Simulator fixes

jmuhlich:
Rebased version of #192 from @JamesPino. Original comment:

> Corrected an lsoda error that occured if _use_analytic_jacobian was False. Before it would try to assign self.jac_fn = lambda t, y, p: jacobian(y, t, p). But jacobian is only defined if _use_analytic_jacobian =True.
> 
> The rest are enhancments to the simulator class. A lot of calls could be done once outside of the main loop, which sped up the _calc_yobs_yexp quite a bit.

This still ought to have a test that would fail before the jac_fn fix. Will implement that now.

### DIFF
--- a/pysb/integrate.py
+++ b/pysb/integrate.py
@@ -279,14 +279,24 @@ class Solver(object):
 
         # Integrator
         if integrator == 'lsoda':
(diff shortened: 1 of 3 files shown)


johnbachman:
:+1:

PR #196, opened 2016-04-01
Removed simulate.py for future reintroduction

jmuhlich:
@lh64 and I agreed this isn't ready for prime time yet and we don't want to commit to an API before we've implemented more subclasses. It will live on in some of the LoLab branches as needed until we decide to reintroduce it here.


lh64:
:+1:

Issue #83, opened 2013-08-09
pysb.kappa shoud be more intelligent about looking for likely locations of kappa


johnbachman:
We should crib from bng and look for likely install locations or allow the setting of an environment variable.

jmuhlich:
John, we should really do this for v1.1.0. Can you take it on?

johnbachman:
Any suggestions for where to look? Users can download and build KaSim
anywhere.

jmuhlich:
Copy the bng.py BNGPATH handling code, and more or less just change BNGPATH to KAPPAPATH? Not sure if you should leave the default paths (/usr/local/share... and c:\program files\...) or not.

jmuhlich:
Fixed in #200

PR #200, opened 2016-04-05
Kappa path environment variables and directory search

johnbachman:
This pull request is modeled after the approach to finding the path to the BNG executable with a couple of modifications. In particular, while it assumes that the KaSim and KaSa executables are in the same directory, if the user sets the kappa path using set_kappa_path it errors if KaSim is not found in the given directory but only warns if KaSa is not found. This is because KaSa is not actually included with the (current) Kappa binary downloads, so we don't want to prevent people from running KaSim just because they haven't built KaSa.

If the user sets the path to Kappa using the KAPPAPATH environment variable, KaSim and KaSa are expected to both reside in that location, but if KaSa is missing the code only errors if the user attempts to run one of the static analysis functions (run_static_analysis, contact_map, or influence_map).

This should take care of issue #83. @jmuhlich, let me know what you think.

### DIFF
--- a/pysb/kappa.py
+++ b/pysb/kappa.py
@@ -40,34 +40,29 @@ def _check_dist_dir(kappa_prog, dist_dir):
     else:
         return None
 
-def set_kappa_path(dir):
+def set_kappa_path(path):
     """Set the path to the KaSim and KaSa executables.
 
     Parameters
     ----------
-    dir : string
+    path: string
         Directory containing KaSim and Kasa executables.
     """
 
     global _kasim_path
     global _kasa_path
     # Check locations of both KaSim and KaSa
-    kasim_path = _check_dist_dir('KaSim', dir)
-    kasa_path = _check_dist_dir('KaSa', dir)
-
-    # Check that KaSim executable exists and is not a directory
+    kasim_path = _check_dist_dir('KaSim', path)
+    kasa_path = _check_dist_dir('KaSa', path)
+    # KaSim is required, so raise if we didn't find it.
     if kasim_path is None:
-        raise Exception('Could not find KaSim in %s.' % os.path.abspath(dir))
-    else:
-        _kasim_path = kasim_path
-
-    # Check KaSa, but only raise a warning if it's not found because it's
-    # not installed by default with the KaSim distribution.
+        raise Exception('Could not find KaSim in %s.' % os.path.abspath(path))
+    # KaSa is optional, so only warn if we didn't find it.
     if kasa_path is None:
         warnings.warn('Could not find KaSa in %s; install KaSa for static '
-                      'analysis features.' % os.path.abspath(dir))
-    else:
-        _kasa_path = kasa_path
+                      'analysis features.' % os.path.abspath(path))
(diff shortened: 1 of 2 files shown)


pysb/kappa.py
```diff
@@ -30,6 +29,99 @@
...
+_kasa_path = None
+
+def _check_dist_dir(kappa_prog, dist_dir):
+    kappa_prog_path = os.path.join(dist_dir, kappa_prog)
+    if os.access(kappa_prog_path, os.F_OK):
+        return kappa_prog_path
+    else:
+        return None
+
+def set_kappa_path(dir):
```
jmuhlich: Don't shadow `dir`. I know the original code from bng.py does it, but that should be fixed too!

pysb/kappa.py
```diff
@@ -30,6 +29,99 @@
...
+        Directory containing KaSim and Kasa executables.
+    """
+
+    global _kasim_path
+    global _kasa_path
+    # Check locations of both KaSim and KaSa
+    kasim_path = _check_dist_dir('KaSim', dir)
+    kasa_path = _check_dist_dir('KaSa', dir)
+
+    # Check that KaSim executable exists and is not a directory
```
jmuhlich: Comment is inaccurate. The check was already done above, and it doesn't check that the full path is not a directory. The "not a directory" bit can probably just be dropped entirely, anyway.

pysb/kappa.py
```diff
@@ -30,6 +29,99 @@
...
+        else:
+            raise Exception('Could not find %s installed in one of the '
+                            'following locations:' % kappa_prog +
+                            ''.join('\n    ' + d for d in dist_dirs) +
+                            '\nSet KAPPAPATH environment variable to '
+                            'directory containing KaSim and KaSa.')
+    # Cache path for future use
+    if kappa_prog == 'KaSim':
+        _kasim_path = kappa_prog_path
+    else:
```
jmuhlich: The `else` should probably be `elif kappa_prog == 'KaSa'` just like on line 90.

jmuhlich:
Looks good other than a few style issues I commented on. Tested locally and verified the Exception works as expected.

pysb/kappa.py
```diff
@@ -30,6 +29,99 @@
...
+    dist_dirs = [
+        '/usr/local/share/KaSim',
+        'c:/Program Files/KaSim',
+        ]
+
+    # First check the environment variable, which has the highest precedence
+    if path_var in os.environ:
+        kappa_prog_path = _check_dist_dir(kappa_prog, os.environ[path_var])
+        if not kappa_prog_path:
+            raise Exception('Environment variable %s is set to but %s could'
```
jmuhlich: Looks like you were going to include the value of `$KAPPAPATH` after 'to' here but forgot to finish.

jmuhlich:
I am going to tackle these and send you a PR to your own branch. :) You can merge-with-squash (a new thing on github) then we can merge this PR the normal way.

johnbachman:
@jmuhlich and I reviewed after merging his fixes.

:+1:

PR #201, opened 2016-04-11
Update installation section of docs and sphinx configuration

jmuhlich:
This overhauls the dependencies section to be more accurate and concise, and also includes instructions for where to install BNG and Kappa (including the BNGPATH and KAPPAPATH environment variables). I realize now that the KAPPAPATH bit should actually be part of #200, but since that's about to be merged I don't think there's any problem.

I also updated the sphinx conf.py to not error in the face of our new Versioneer-powered versioning system, and brought it up to date with recent versions of sphinx which have deprecated the pngmath extension in favor of imgmath.


johnbachman:
Nice :+1:

PR #204, opened 2016-04-22
BioNetGen installation script, with setuptools --install-bng option

alubbock:
This PR adds a script for downloading and installing BioNetGen into the
platform-specific location expected by PySB (or any other specified location).
This option reduces hassle for new users, facilitates cross-platform unit
testing (by installing the correct BNG binary for the current platform) and 
moves us closer to a one step install:
`pip install --install-option="--install-bng" pysb`.

Untested on Windows as yet.


alubbock:
Closing this PR - Anaconda seems a better way to go. See #324.

PR #206, opened 2016-05-16
BioNetGen .bngl model file import for PySB

### LINKED ISSUE #207: Compartmental models with species synthesis/degradation cause an error
`enable_synth_deg()` fails for compartmental models, as it tries to add a `__source()` initial without a compartment, leading to `InvalidInitialConditionError: Pattern not concrete` when one tries `pysb.generator.bng.BngGenerator(model).get_content()`.

### LINKED ISSUE #208: BngConsole errors when getting console output on Python 3
When using the `BngConsole` in Python 3 with the `verbose=True` option, or when a warning or error is generated, Python attempts to read `self.console.before`, which is type `bytes` in Python 3 rather than `str`. It needs to be explicitly decoded using `mystr.decode('utf-8')`. Partial example traceback below.

```python
...
  File "/Users/alex/git/VU/pysb/pysb/importers/bngl.py", line 29, in __init__
    with BngConsole(model=model, verbose=True) as con:
  File "/Users/alex/git/VU/pysb/pysb/bng.py", line 297, in __init__
    raise BngInterfaceError(e)
pysb.bng.BngInterfaceError: 'str' does not support the buffer interface
```

### LINKED ISSUE #209: `ComplexPattern.is_equivalent_to()` fails on Python 3
The `ComplexPattern.is_equivalent_to()` method relies on sorting a tuple containing `(str, dict, int)`. `dict` is no longer sortable in Python 3 without specifying a sort order.

alubbock:
Adds a new tool `pysb.importers.bngl`, which reads in a .bngl file using the BioNetGen Console to convert them to XML, then parsing that XML into a PySB Model object. The big idea here is to give us access to many of the models written in BNGL automatically. In addition, it opens up the possibility of using BioNetGen's [Atomizer](http://bionetgen.org/index.php/SBML2BNGL) to read SBML models into PySB automatically, by doing SBML->BNGL->PySB.

The BNGL import works, but has a few limitations due to PySB lack of support for local functions, monomers containing multiple sites with the same name, and multiple binding to the same site. See documentation for further details. Nevertheless, it should work for most use cases.

Example usage:
```python
from pysb.importers.bngl import model_from_bngl
model = model_from_bngl('/path/to/my.bngl')
```

This PR also fixes #207 in 6e2087f, fixes #208 in f7000d8, and fixes #209 in bd447df.


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: python
-sudo: false
+sudo: required
```
jmuhlich: Do we really need sudo or did this just slip in by accident?

pysb/bng.py
```diff
@@ -103,12 +104,16 @@ class BngBaseInterface(object):
...
     def __init__(self, model, verbose=False, cleanup=False,
                  output_prefix=None, output_dir=None):
+        self._base_file_stem = 'pysb'
         self.verbose = verbose
         self.cleanup = cleanup
         self.output_prefix = 'tmpBNG' if output_prefix is None else \
             output_prefix
-        self.generator = BngGenerator(model)
-        self._check_model()
+        if model:
```
jmuhlich: Is there a use case for calling the constructor with model=None? A lot of the code in this class can be simplified if model is guaranteed to be defined.

pysb/core.py
```diff
@@ -1426,8 +1427,15 @@ def enable_synth_deg(self):
             self.add_component(Parameter('__source_0', 1.0, _export=False))
 
         source_cp = as_complex_pattern(self.monomers['__source']())
-        if not any(source_cp.is_equivalent_to(other_cp) for other_cp, value in self.initial_conditions):
-            self.initial(source_cp, self.parameters['__source_0'])
+        if self.compartments:
+            for c in self.compartments:
+                source_cp.compartment = c
```
jmuhlich: This is a bug. Because it modifies source_cp in place, on the second loop it ends up altering the initial condition added in the first loop, making that one match the current proposed source_cp, and skipping it. This keeps happening until the end of the loop, at which point there's only one \__source initial condition for the last compartment in the model. source_cp needs to be copied on each loop, and the easiest way is just to use the proper `**` operator to refine it against the current compartment. Just replace this line with `source_cp = source_cp ** c`.

pysb/importers/bngl.py
```diff
@@ -0,0 +1,319 @@
...
+                raise BnglImportError('Parameter %s has unknown type: %s' %
+                                      (p_name, p.get('type')))
+
+    def parse_observables(self):
+        for o in self.x.iterfind(_ns('{0}ListOfObservables/{0}Observable')):
+            o_name = o.get('name')
+            cplx_pats = []
+            for mp in o.iterfind(_ns('{0}ListOfPatterns/{0}Pattern')):
+                match_once = mp.get('matchOnce')
+                match_once = 1 if match_once is not None \
```
jmuhlich: `is not None` is redundant with `== 1`.

pysb/importers/bngl.py
```diff
@@ -0,0 +1,319 @@
...
+                                                compartment=None,
+                                                match_once=match_once))
+            self.observable(o_name,
+                            ReactionPattern(cplx_pats),
+                            match=o.get('type').lower())
+
+    def parse_initials(self):
+        for i in self.x.iterfind(_ns('{0}ListOfSpecies/{0}Species')):
+            if i.get('Fixed') is not None and i.get('Fixed') == "1":
+                warnings.warn('Species %s is fixed, but will be treated as '
```
jmuhlich: Is it reasonable to warn and continue in this case, or should it be fatal? The kinetics of the resulting model is going to be drastically different.

pysb/importers/bngl.py
```diff
@@ -0,0 +1,319 @@
...
+            rate_law_name = rl.find(_ns('{0}ListOfRateConstants/'
+                                        '{0}RateConstant')).get('value')
+            try:
+                return self.model.parameters[rate_law_name]
+            except KeyError:
+                return self.model.expressions[rate_law_name]
+        elif rl.get('type') == 'Function':
+            return self.model.expressions[rl.get('name')]
+        else:
+            raise BnglImportError('Rate law %s has unknown type %s' %
```
jmuhlich: Shouldn't we be able to implement `MM` and `Sat` with Expressions? It's fine to punt on it for now but it seems like we could handle these if we wanted to.

pysb/importers/bngl.py
```diff
@@ -0,0 +1,319 @@
...
+                elif bonds == "1":
+                    bond_list = bond_ids[comp.get('id')]
+                    assert int(bonds) == len(bond_list)
+                    mon_states[state_nm] = bond_list[0]
+                elif bonds == "?":
+                    mon_states[state_nm] = WILD
+                elif bonds == "+":
+                    mon_states[state_nm] = ANY
+                else:
+                    raise BnglImportError('Bond %s on monomer %s has '
```
jmuhlich: Pysb handles multi-bonds just fine. The only place they cause trouble is Kappa export (at which point an exception is thrown if any exist). Is there another reason I'm missing that this is a problem here?

jmuhlich:
I think the new `pysb.importers` should be renamed `pysb.import` for symmetry and consistency with `pysb.export`.

pysb/importers/bngl.py
```diff
@@ -0,0 +1,319 @@
...
+    Shorthand to apply XML namespace
+    """
+    return tag_string.format('{http://www.sbml.org/sbml/level3}')
+
+
+class BnglImportError(Exception):
+    pass
+
+
+class BnglBuilder(Builder):
```
jmuhlich: All of the methods and properties introduced in this subclass should really be underscore-prefixed, no? I know we haven't been very consistent with this sort of thing in the past, but with SemVer I am getting a lot more sensitive to exposing things as public by accident. The public interface to this class should probably just be the new constructor and all the public interface from Builder.

.travis.yml
alubbock: This is needed - the [docs](https://docs.travis-ci.com/user/trusty-ci-environment/) state "all Trusty builds have sudo enabled".

pysb/bng.py
alubbock: A limited one: using the BNG interface with a model loaded directly from a BNGL file, as I've done in the new unit tests for this PR in pysb/tests/test_importers.py. It does complicate the code slightly but the added functionality is useful, unless I'm missing another way of achieving this.

pysb/importers/bngl.py
```diff
@@ -0,0 +1,319 @@
...
+    """
+    Class for assembling a Model from a .bngl file. See
```
jmuhlich: First line of a docstring is supposed to be just one line. The details go in the body, after a blank line. Also I think "Class for..." is redundant. Perhaps just "Assemble a Model from a .bngl file".

pysb/importers/bngl.py
```diff
@@ -0,0 +1,319 @@
...
+        self.parse_compartments()
+        self.parse_initials()
+        self.parse_observables()
+        self.parse_expressions()
+        self.parse_rules()
+
+
+def model_from_bngl(filename):
+    """
+    Convert a BioNetGen .bngl model definition file into a PySB Model
```
jmuhlich: Same as above, summary should be a single line.

pysb/importers/sbml.py
```diff
@@ -0,0 +1,129 @@
...
+                    output_file=None,
+                    convention_file=None,
+                    naming_conventions=None,
+                    user_structures=None,
+                    molecule_id=False,
+                    atomize=False,
+                    pathway_commons=False,
+                    verbose=False):
+    """
+    Runs the BioNetGen sbmlTranslator binary. For more descriptions of
```
jmuhlich: Keep summary to one line, and reformat params section using numpydoc/napoleon conventions.

pysb/tests/test_importers.py
```diff
@@ -0,0 +1,129 @@
...
+        bng.action('simulate', method='ode', sample_times=sim_times)
+        yfull2 = bng.read_simulation_results()
+
+    # Check all species trajectories are equal (within numerical tolerance)
+    for species in m.species:
+        print(species)
+        print(yfull1[species])
+        print(yfull2[species])
+        print(numpy.allclose(yfull1[species], yfull2[species]))
+        assert numpy.allclose(yfull1[species], yfull2[species])
```
jmuhlich: Given that this should be the exact same set of ODEs, the default `allclose` tolerance might be too loose! Maybe we don't want to depend on the BNG ODE simulation being deterministic and expect precise equality, but we should at least make the tolerance here as tight as the default BNG solver's tolerance.

pysb/importers/sbml.py
```diff
@@ -0,0 +1,129 @@
...
+    tmpdir = tempfile.mkdtemp()
+    verbose = kwargs.get('verbose', False)
+    if verbose:
+        print("Performing SBML to BNGL translation in temporary "
+              "directory %s" % tmpdir)
+    try:
+        bngl_file = os.path.join(tmpdir, 'model.bngl')
+        sbml_translator(filename, bngl_file, **kwargs)
+        return model_from_bngl(bngl_file)
+    finally:
```
jmuhlich: This silently swallows all exceptions... is that what was intended?

.travis.yml
jmuhlich: I see, they've moved to a fully containerized environment so sudo is not a problem anymore. Odd that they force `sudo: required` though. :+1:

pysb/bng.py
jmuhlich: Oh I see, I hadn't realized you could do that. Perhaps the signature here should have `model=None` then to be more explicit that it's optional. Then in any method where a model is needed, the first thing it does is raise an exception if self.model is None. (Also, why not keep a reference to model in self rather than the property that fetches it from the generator?)

pysb/bng.py
```diff
@@ -236,7 +243,7 @@ def read_simulation_results(self):
         # Read concentrations data
         cdat_arr = numpy.loadtxt(self.base_filename + '.cdat', skiprows=1)
         # Read groups data
-        if len(self.model.observables):
+        if self.model and len(self.model.observables):
```
jmuhlich: So if the model is loaded straight from a .bngl file and simulated, the observable trajectories won't be read in? Or am I reading this wrong?

pysb/importers/bngl.py
alubbock: Unimplemented functionality which changes model behaviour should probably error by default, yes. It'd be good to have a `force=True` option to build the model partially anyway, so that advanced users can fix it with manual edits.

pysb/importers/bngl.py
alubbock: I think @lh64 said these were deprecated in BNG, but having a few canned rate law expressions would nonetheless be useful to have in a future release.

alubbock:
Re @jmuhlich 
> I think the new pysb.importers should be renamed pysb.import for symmetry and consistency with pysb.export.
This causes problems as `import` is a reserved Python keyword.

pysb/importers/bngl.py
jmuhlich: That sounds like a reasonable approach.

jmuhlich:
Sigh... pysb.importers it is, I guess. :unamused:

pysb/importers/sbml.py
alubbock: In this case, no exceptions are caught; they are handled as if the `try`/`finally` block wasn't there. Catching and suppressing exceptions would require an `except: pass` block.

As an example, the following raises an Exception:

```python
try:
    raise Exception()
finally:
    print('finally')
```

pysb/bng.py
alubbock: Implemented in 3db22ab

pysb/importers/bngl.py
alubbock: Implemented in 2fd9d8a

pysb/core.py
alubbock: Implemented in 02e3204

pysb/importers/bngl.py
alubbock: Implemented in 7cf3c2e

pysb/importers/bngl.py
alubbock: Implemented in 9c9a15f

pysb/importers/bngl.py
alubbock: Implemented in 9620729

pysb/importers/bngl.py
alubbock: Implemented in 2fd9d8a

pysb/importers/bngl.py
alubbock: Implemented in 2fd9d8a

pysb/importers/sbml.py
alubbock: Implemented in 2fd9d8a

pysb/tests/test_importers.py
alubbock: Implemented in 757e2f2

pysb/bng.py
alubbock: You're right, observables are not read in if `model=None`. The use case for `model=None` is limited (probably only internal unit tests?), but in the future best practice would probably be to check the header of the observables file and verify those match the model. Expressions should be read in from there too, to save re-computing them in PySB.

PR #212, opened 2016-06-07
Simulator Class

alubbock:
This is a new Simulator class, designed to replace and extend the old Solver class and provide a consistent interface across all simulators. For now, only the SciPy ODE solvers are implemented, in the ScipyOdeSimulator class.

There are basically two modes of operation: `Simulator.execute()`, a class method which works like the old `odesolve` method in that it is a one-liner to execute a simulation, and `Simulator.run()`, an instance method which is more efficient when running multiple simulations in sequence, e.g. with different initial conditions.

Trajectories are accessed via the methods `concs_species()`, `concs_observables()`, `concs_expressions()`, and `concs_all()` (as `execute()` returns) which includes species, expressions and observables together in a single array. (concs=concentrations, in case that wasn't obvious).

The class also allows per-simulation manual override of model initial conditions or parameters, both of which can now be supplied as a list (to change everything) or a dictionary (to change specific initials/parameters).


pysb/integrate.py
```diff
@@ -101,207 +61,27 @@ class Solver(object):
...
-            # Initialization of matrix for storing the Jacobian
-            self.jac = numpy.zeros((len(self.model.odes), len(self.model.species)))
-            jac_fn = jacobian
-
-        # build integrator options list from our defaults and any kwargs passed to this function
-        options = {}
-        if default_integrator_options.get(integrator):
-            options.update(default_integrator_options[integrator]) # default options
+    @property
+    def _use_inline(self):
```
jmuhlich: This should need a setter to pass through writes as well.

pysb/simulator/__init__.py
```diff
@@ -0,0 +1 @@
+from scipyode import ScipyOdeSimulator
```
jmuhlich: Should be a relative import -- `from .scipyode ...`. Otherwise you'll get an error on Python 3 due to absolute_import being in effect.

pysb/simulator/base.py
```diff
@@ -0,0 +1,329 @@
...
+
+    """
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def __init__(self, model, verbose=False, **kwargs):
+        self.model = model
+        self.verbose = verbose
+        self.tout = None
+        self._y = None
```
jmuhlich: Maybe just call self._clear_outputs instead of duplicating all those `= None` lines here? _clear_outputs might better be renamed to _init_outputs then.

pysb/simulator/base.py
```diff
@@ -0,0 +1,329 @@
...
+        self._yobs = None
+        self._yobs_view = None
+        self._yexpr = None
+        self._yexpr_view = None
+        self._yfull = None
+        # Per-run initial conditions/parameter/tspan override
+        self.tspan = kwargs.get('tspan', None)
```
jmuhlich: Did you avoid making tspan, initials and param_values explicit named args in the method signature so the signature wouldn't be unwieldy for subclasses? I noticed that initials and param_values aren't documented in the class docstring. Does it make sense to allow passing these three things to the constructor, or should we only support setting them on the object after initialization is complete?

pysb/simulator/base.py
```diff
@@ -0,0 +1,329 @@
...
+            return self.initials_list
+
+    @initials.setter
+    def initials(self, new_initials):
+        if new_initials is None:
+            self._initials = None
+        else:
+            # check if y0 is a dict, and if so validate the keys
+            # (ComplexPatterns)
+            if isinstance(new_initials, dict):
```
jmuhlich: Should probably check against `collections.Mapping` rather than `dict`.

pysb/simulator/base.py
```diff
@@ -0,0 +1,329 @@
...
+                                                 ComplexPattern)):
+                        raise SimulatorException('Dictionary key %s is not a '
+                                                 'MonomerPattern or '
+                                                 'ComplexPattern' %
+                                                 repr(cplx_pat))
+                self._initials = new_initials
+            # accept vector of species amounts as an argument
+            elif len(new_initials) != self._y[0].shape[1]:
+                raise ValueError("y0 must be the same length as model.species")
+            elif not isinstance(new_initials, np.ndarray):
```
jmuhlich: You can replace the last elif and the else with this:
```python
else:
    self._initials = np.array(new_initials, copy=False)
```
That will construct a new array for a non-array, and return a view on an existing array (shares the same memory and avoids a copy).

pysb/simulator/base.py
alubbock: The original motivation was to make subclassing easier (perhaps some subclasses would not support setting initials in the constructor, for example). But the overhead for implementing this in a subclass seems low, so I've changed the base class to allow setting tspan, initials and param_values in the constructor.

alubbock:
@jmuhlich Thanks for the feedback on the Simulator class, which I've now implemented.

jmuhlich:
There's something wrong with the commit history -- I think this branch was (correctly) rebased against master and later (incorrectly) merged with the original. This introduces lots of "duplicate" commits in the parallel branches. I think the latest commit 7933a0d3dc791e67d63373d3e0eada33cff84e24 on `alubbock/simulatorclass` is the suspect merge. I think you want to do `reset --hard f95d7e20e9cfc3ce278abc3323538da94585c479` on your branch and then `push -f`. Of course double-check that before pulling the trigger. :)

jmuhlich:
It's also odd that the commit message for that merge doesn't look like the standard git merge commit message but rather looks like a normal manually entered commit message: "Update Tyson oscillator example to use Simulator class". This is another clue that something went wrong.

alubbock:
Hmm.. not entirely sure why I did that! Apologies for the mistake. I've reset the merge as you suggested.

jmuhlich:
Thanks. I thought about the big picture of the API this time rather than just scrutinizing the low-level details. Some thoughts:
- I suspect some will confuse `execute` and `run`, and try to call `execute` on an instance. Can we think of a less confusing name for `execute`, and arrange to throw an exception if it's called on an instance? I think it will require assigning a top-level function to an attribute of the class rather than using `@classmethod` on a regular method.
- The `concs_*` methods need a different naming scheme, as not all simulators return concentrations (the stochastic ones return integer molecule counts, for example). Maybe `traj_*` for "trajectories"? Also they could be getters for ease of use.
- Furthermore, would it be cleaner to just have a standalone "simulation result" object rather than put all these methods directly on the Simulator object? I think that would provide a more natural interface for users -- `run()` could actually return the result rather than just causing a side effect. `execute` (or whatever it ends up being named) could return one of these objects as well.

jmuhlich:
I did some testing and came up with this:

```python
def _Foo_instance_func(*args, **kwargs):
    raise ValueError("func may only be called on subclasses of Foo, not instances.")
    
class Foo(object):
    def __init__(self):
        self.func = _Foo_instance_func
    @classmethod
    def func(cls, a):
        print "cls=%s, a=%s" % (cls, a)

class Bar(Foo):
    pass

f = Foo()
b = Bar()
f.func()        # raises
f.func('abc')   # raises
f.func(x=123)   # raises
Foo.func(1)     # prints: cls=<class '__main__.Foo'>, a=1
b.func()        # raises
Bar.func(2)     # prints: cls=<class '__main__.Bar'>, a=2
```

alubbock:
Thanks again for your feedback @jmuhlich. As suggested, I've added a `SimulationResult` class. This also solves both the `concs_` naming prefix issue (since the methods are now simply properties of the `SimulationResult`, like `result.expressions` and `result.observables`), and the `execute()` vs `run()` issue (now that `run()` returns something, it allows simulations as a one-liner thus making `execute()` at the class level redundant).

I also added the option to return simulation results in a pandas DataFrame with a MultiIndex, and fixed a travis build issue on python 3.4 (see #229) in the other two commits respectively.

pysb/simulator/base.py
```diff
@@ -307,30 +281,27 @@ def concs_all(self, squeeze=True):
...
-        if self.n_sims == 1 and squeeze:
+        if self.nsims == 1 and self.squeeze:
             return self._yfull[0]
         return self._yfull
 
-    def concs_species(self, squeeze=True):
-        if self.n_sims == 1 and squeeze:
+    @property
+    def species(self):
+        if self.nsims == 1 and self.squeeze:
```
jmuhlich: :+1: on `squeeze`! I missed this feature in the original code.

pysb/simulator/base.py
```diff
@@ -190,81 +179,67 @@ def param_values(self, new_params):
...
+class SimulationResult(object):
+    """
+    Contains the results of a simulation, combined with properties and
+    methods to access them.
+    """
+    def __init__(self, simulator, trajectories):
+        self.squeeze = True
+        self.simulator = type(simulator).__name__
+        self.tout = simulator.tout
+        self._y = trajectories
```
jmuhlich: Probably want to raise a `ValueError` if `trajectories.ndim != 3`. If we get other folks implementing simulators, I can imagine that everyone will pass a 2D array for a single simulation in their initial implementation. Let's catch that early to aid in diagnosis.

jmuhlich: Should we also check that the appropriate dimension of `trajectories` (the last one, I think) is compatible with the length of `tout`? I think that's invariant across all possible simulators.

alubbock: `trajectories` can either be a 3D array or a list of 2D arrays (in the case where `tspan` is different for each simulation). I'll add some validation checks for each case.

jmuhlich:
I just played with the API a bit and I like it. The `dataframe` method is an especially nice touch.

pysb/simulator/base.py
```diff
@@ -0,0 +1,324 @@
...
+            return self._yfull[0]
+        return self._yfull
+
+    @property
+    def dataframe(self):
+        if pd is None:
+            raise Exception('Please "pip install pandas" for this feature')
+        sim_ids = (np.repeat(range(self.nsims), [len(t) for t in self.tout]))
+        times = np.concatenate(self.tout)
+        idx = pd.MultiIndex.from_tuples(zip(sim_ids, times),
```
jmuhlich: Oops, this should honor `squeeze`!

pysb/simulator/base.py
alubbock: Implemented in a40236f

pysb/simulator/base.py
alubbock: Implemented in aca4263

pysb/simulator/base.py
alubbock: Implemented in b4a8a0c

PR #213, opened 2016-06-10
Access model components as attributes

alubbock:
`alias_model_components()` is a useful function that @lh64 and I think should go
in core. In keeping with the Python idiom "explicit is better than implicit",
I've added an argument to alias model components by name, or by a specific
component type. The latter is useful as often only the monomers are needed,
e.g. to construct ComplexPatterns for specifying override initial conditions
in the new Simulator class, or for extending existing models in a separate
Python module.


alubbock:
Worth considering whether we want to change the `caller_globals.update()` line. Any existing symbols in the caller's global namespace will be silently overwritten. Would raising an exception in that case be preferable?

pysb/core.py
```diff
@@ -1439,6 +1439,55 @@ def reset_equations(self):
...
+                                                           's'))
+        else:
+            all_components = self.all_components()
+            component_names = set(c.name for c in all_components)
+            missing_components = [nm for nm in names_or_type if nm not in
+                                  component_names]
+            if missing_components:
+                raise ValueError('Components not found in model: ' + repr(
+                    missing_components))
+            components = dict((c.name, c) for c in all_components if c.name
```
jmuhlich: I think this would be clearer written something like this:
```python
components = {n: all_components[n] for n in names_or_type}
```

pysb/util.py
```diff
@@ -9,7 +9,13 @@
 
 
 def alias_model_components(model=None):
-    """Make all model components visible as symbols in the caller's global namespace"""
+    """
```
jmuhlich: This ought to generate a `DeprecationWarning`, e.g.
```python
warnings.warn("alias_model_components is deprecated, please use Model.alias_components",
              DeprecationWarning, stacklevel=2)
```
Python does silence that warning by default but it does seem to be the proper thing to do.

Also the docstring should use a sphinx deprecation notice rather than plain text:
```rst
.. deprecated:: 1.2
   Use :func:`pysb.core.Model.alias_components` instead.
```

jmuhlich:
The more I think about this, the more I dislike it. What if we implement a special object contained in the Model that implements `__getattribute__` as a pass-through to Model.all_components.__getitem__ ? Let's call it "comps". Then you could access e.g. monomer "A" as `model.comps.A` and calling alias_model_components could be replaced by doing `c = model.comps`  and referencing `c.A` rather than just `A`. Thoughts? @johnbachman feel free to chime in too.

jmuhlich:
After having used Pandas a lot, I feel comfortable with its approach to accessing DataFrame columns (Series) which I think we could adopt:

* A column with a "clean" name (matches `[A-Za-z_][A-Za-z0-9_]*`) that doesn't conflict with an actual DataFrame attribute or method can be accessed as `df.foo`.
* Any column can be accessed as `df['foo']` no matter what its name looks like or whether it shares a name with a DataFrame attribute or method.
* Iterating over the DataFrame yields its columns.

Now read that again, replacing "column" with "component", and "DataFrame" with "model". I think we could also give ComponentSet the same behavior (in addition to its existing \_\_getitem__ implementation) to support things like `model.monomers.foo`.

jmuhlich:
Oops, I forgot that iterating over a DataFrame yields column names, not the Series themselves. I think we could iterate over Components rather than their names if we implement this.

alubbock:
I've implemented @jmuhlich's suggestion of allowing access to 
model components as attributes, as well as items. This replaces
the previous PR, which I've force-pushed over. There's also
a new model.comps property, which gets all model 
components. Thus, we can do things like:

```python
model.monomers.Bax    # Get the Monomer 'Bax'
model.comps.rule_cat  # Get the Rule 'rule_cat'
model.parameters.kf1  # Get the Parameter 'kf1'
model.comps.kf1       # Ditto
```

This should help when importing models, and reduce the need for
`alias_model_components()`, which can cause namespace issues,
particularly when we want access to components from more than
one model.

In actual use, one could alias `model.monomers` for easy access,
e.g.

```python
import model1
import model2

m1 = model1.monomers
m2 = model2.monomers

m1.Bax  # Bax from model 1
m2.Bax  # Bax from model 2
```

pysb/core.py
```diff
@@ -1643,6 +1647,14 @@ def __getitem__(self, key):
         else:
             return self._map[key]
 
+    def __getattr__(self, name):
+        if name == '_map':
```
jmuhlich: What is this guarding against?

pysb/core.py
```diff
@@ -1643,6 +1647,14 @@ def __getitem__(self, key):
...
+            raise AttributeError()
+        try:
+            return self[name]
+        except KeyError as e:
+            raise AttributeError(e)
```
jmuhlich: The standard AttributeError message is of the form `'Foo' object has no attribute 'bar'`. This code just produces a message of the form `'bar'`. Maybe the message should be something like `Model has no component 'bar'`, e.g. 
```python
raise AttributeError("Model has no component '%s'" % name)
```

pysb/tests/test_core.py
```diff
@@ -4,10 +4,31 @@
...
+        assert not hasattr(model.comps, name)
+        # Add the element to a model and try to access it as attribute and item
+        model.add_component(c)
+        assert_equal(model.comps[name], c)
+        assert_equal(getattr(model.comps, name), c)
+
+
+@with_model
+def test_component_name_existing_attribute():
+    # _map is an attribute of ComponentSet, the superclass of Monomer
```
jmuhlich: ComponentSet is not the superclass of Monomer... These two comment lines can probably just be deleted, since the comment before the asserts explains the purpose.

pysb/core.py
alubbock: Infinite recursion. Without this check, `self._map[key]` in `__getitem__` calls `__getattr__`, which calls `__getitem__`.....

pysb/core.py
alubbock: Implemented.

pysb/tests/test_core.py
alubbock: Removed the misleading comment as requested.

pysb/core.py
jmuhlich: Did you actually witness the recursion? I'm pretty sure `__getattr__` is only called for attributes that don't actually exist in the object. Maybe you were thinking of `__getattribute__` which is called unconditionally?

alubbock: Yes, I did witness it. Commenting out the `if` clause gives the following:

```
$ nosetests pysb.tests.test_core
.........E.................
======================================================================
ERROR: pysb.tests.test_core.test_model_pickle
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/alex/.virtualenvs/pysb/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/alex/git/VU/pysb/pysb/testing/__init__.py", line 18, in inner
    func(*args, **kwargs)
  File "/Users/alex/git/VU/pysb/pysb/tests/test_core.py", line 112, in test_model_pickle
    model2 = pickle.loads(pickle.dumps(model))
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 1388, in loads
    return Unpickler(file).load()
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 864, in load
    dispatch[key](self)
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 1221, in load_build
    setstate = getattr(inst, "__setstate__", None)
  File "/Users/alex/git/VU/pysb/pysb/core.py", line 1839, in __getattr__
    return self[name]
  File "/Users/alex/git/VU/pysb/pysb/core.py", line 1833, in __getitem__
    return self._map[key]
  File "/Users/alex/git/VU/pysb/pysb/core.py", line 1839, in __getattr__
    return self[name]
  File "/Users/alex/git/VU/pysb/pysb/core.py", line 1833, in __getitem__
    return self._map[key]
  File "/Users/alex/git/VU/pysb/pysb/core.py", line 1839, in __getattr__
    return self[name]
  File "/Users/alex/git/VU/pysb/pysb/core.py", line 1833, in __getitem__
    return self._map[key]
...<snip>...
RuntimeError: maximum recursion depth exceeded
```

jmuhlich: Oh, that looks like a pickling-specific problem. A little googling for `__getattr__` and pickling led to this bug report and the resulting section added to the docs:

https://bugs.python.org/issue5370
https://docs.python.org/2.7/library/pickle.html#pickling-and-unpickling-normal-class-instances

> At unpickling time, some methods like `__getattr__()`, `__getattribute__()`, or `__setattr__()` may be called upon the instance. In case those methods rely on some internal invariant being true, the type should implement either `__getinitargs__()` or `__getnewargs__()` to establish such an invariant; otherwise, neither `__new__()` nor `__init__()` will be called. 

Can we fix this at the level of the pickle protocol? The current solution is confusing to me.

pysb/importers/sbml.py
```diff
@@ -15,7 +15,7 @@
 from pysb.logging import get_logger, EXTENDED_DEBUG
 
 BIOMODELS_REGEX = re.compile(r'(BIOMD|MODEL)[0-9]{10}')
-BIOMODELS_URL = 'http://www.ebi.ac.uk/biomodels-main/download?mid={}'
+BIOMODELS_URL = 'http://biomodels.caltech.edu/download?mid={}'
```
jmuhlich: Looks like this slipped in accidentally.

pysb/core.py
```diff
@@ -1508,6 +1508,10 @@ def all_components(self):
             cset_all |= cset
         return cset_all
 
+    @property
+    def comps(self):
```
jmuhlich: The more I think about this, the more I think it should be spelled out in full as `components`. If someone is using it frequently enough that the name's length is cumbersome, then even `comps` would be too long. They would want to alias it with a local variable with as short a name as possible.

alubbock: Rationale for model.components makes sense. Implemented.

PR #216, opened 2016-06-14
Add MatchOnce to pysb_flat exporter

alubbock:
This is needed by some models, including certain versions of EARM, as requested by @JamesPino.


alubbock:
Fixes lolab-vu/pysb#35.

jmuhlich:
Doh, good catch.

PR #223, opened 2016-06-30
Wrap model.name with str() in case it is None

johnbachman:
If model.name is None then os.path.join will fail. Wrapping in str()
fixes this problem.


jmuhlich:
:+1: waiting for Travis.

jmuhlich:
I just merged another PR so please rebase!

PR #230, opened 2016-09-26
Replace type(s) == str with isinstance(s, basestring)

johnbachman:
There were a few places in core where sites and states of monomers were checked for validity by checking to see if they are strings with `type(s) == str`. However, in my calling code (which I am rewriting to use a Unicode sandwich and be Python 3 compatible) I am passing in Monomer definitions with unicode definitions. This works for Python 3,  in which unicode strings are `str`, but in Python 2, these `type(s) == str` checks fail.

In the update to make pysb Python 3 compatible, core.py already switches to `basestring` checks in most places, with `basestring = str` added so that there is no NameError in Python 3. But these `str` checks have persisted.

In this PR I've updated the checks in the various places to use `isinstance(s, basestring)` instead of `type(s) == str`. I've also moved to `isinstance` everywhere instead of explicit type checks. I also couldn't help wrapping a few lines to 80 chars as I saw them.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -240,10 +240,12 @@ class Monomer(Component):
         if site_states is None:
             site_states = {}
 
-        # ensure sites is some kind of list (presumably of strings) but not a string itself
-        if not isinstance(sites, collections.Iterable) or isinstance(sites, basestring):
+        # ensure sites is some kind of list (presumably of strings) but not a
+        # string itself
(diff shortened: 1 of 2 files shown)


johnbachman:
Added in the regression test, which failed (as it should) on the base branch under Python 2.7.

jmuhlich:
I vouch that @alubbock's review request is resolved.

PR #232, opened 2016-10-08
Documentation updates & new installation instructions

alubbock:
This PR contains new PySB installation instructions for Anaconda and Docker-based installs, which work on Windows, Mac and Linux. Hopefully the new step-by-step approach is clear, but feedback is welcome. I removed the virtual appliance instructions since it's quite out of date now (based on PySB 0.1.9).

I've improved documentation for the Simulation class and SBML/BNGL importers, which are now included in Sphinx output. I also fixed a couple of Sphinx warnings when parsing the `bind_complex` macro docstring and missing pysb_flat exporter docs.

### DIFF
--- a/pysb/simulator/base.py
+++ b/pysb/simulator/base.py
@@ -17,8 +17,9 @@ class SimulatorException(Exception):
 class Simulator(object):
     """An abstract base class for numerical simulation of models.
 
-    Please note that the interface for this class is considered
-    experimental and may change without warning as PySB is updated.
+    .. warning::
+        The interface for this class is considered experimental and may
+        change without warning as PySB is updated.
 
     Parameters
     ----------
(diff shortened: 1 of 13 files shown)


pysb/simulator/base.py
```diff
@@ -213,8 +214,17 @@ class SimulationResult(object):
     """
     Results of a simulation with properties and methods to access them.
 
-    Please note that the interface for this class is considered
-    experimental and may change without warning as PySB is updated.
+    .. warning::
+        Please note that the interface for this class is considered
+        experimental and may change without warning as PySB is updated.
+
+    .. note::
```
jmuhlich: Also make this a Note section?

pysb/simulator/base.py
```diff
@@ -224,27 +234,72 @@ class SimulationResult(object):
         A set of species trajectories from a simulation. Should either be a
         list of 2D numpy arrays or a single 3D numpy array.
 
-    Attributes
```
jmuhlich: Where did the Attributes section go? Does this somehow get auto-populated from the `@property` method docstrings?

alubbock: `@property` method docstrings get listed with methods (see [base_filename from the BNG interface for an example](http://pysb.readthedocs.io/en/latest/modules/bng.html#pysb.bng.BngBaseInterface.base_filename)). Having an Attributes section means they get listed twice and documentation is duplicated. Alternatively, they could all be listed under Attributes, and have a method docstring that says "See Attributes for description". Listing docstrings under each method seems cleaner to me though, and nicer when reading the code directly.

Issue #233, opened 2016-10-11
String/unicode type checking


bgyori:
This errors if a unicode site state is used in Python 2 because `type` is not `str`
https://github.com/pysb/pysb/blob/master/pysb/generator/bng.py#L183
This could have been addressed in #230 but it looks like only core.py was updated there. Maybe there are further files in PySB with this issue.

PR #234, opened 2016-10-12
Fix isinstance(foo, str) in bng and kappa generation

johnbachman:
This is a follow-up to the previous PR (#230) which made the corresponding fix in core.py, and fixes issue #233. We had found that while #230 allowed models to be created with unicode names/sites in Python 2, BNG and Kappa generation would fail due to unicode strings not being matched in the series of tests on site conditions. I've added corresponding tests in test_bng.py and test_kappa.py.

In the interest of completeness, I grepped all PySB source for other code containing `type(foo) == str` or `isinstance(foo, str)` and found cases only in bng.py and (_bng_param method) and render_reactions.py, which I also updated in this pull request.

### DIFF
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -23,6 +23,13 @@ try:
 except ImportError:
     pass
 
+# Python 2
+try:
+    basestring
+# Python 3
+except NameError:
+    basestring = str
+
 # Cached value of BNG path
 _bng_path = None
 
@@ -154,7 +161,7 @@ class BngBaseInterface(object):
         param :
             An argument to a BNG action call
         """
-        if isinstance(param, str):
+        if isinstance(param, basestring):
             return '"%s"' % param
         elif isinstance(param, bool):
             return 1 if param else 0
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -23,10 +23,9 @@ try:
 except ImportError:
     pass
 
-# Python 2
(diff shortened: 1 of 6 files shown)


pysb/bng.py
```diff
@@ -23,6 +23,13 @@
 except ImportError:
     pass
 
+# Python 2
```
jmuhlich: Would rather see a single line comment explaning the whole block... the 2 / 3 comments are not helpful for someone who doesn't know why this is being done. Something like, "Alias basestring under Python 3 for forwards compatibility".

pysb/generator/bng.py
```diff
@@ -3,6 +3,12 @@
 import pysb
 import sympy
 
+# Python 2
```
jmuhlich: Same as in /pysb/bng.py

pysb/generator/kappa.py
```diff
@@ -3,6 +3,13 @@
 from re import sub
 import warnings
 
+# Python 2
```
jmuhlich: Same here.

pysb/tools/render_species.py
```diff
@@ -30,6 +30,13 @@
 import pygraphviz
 import pysb.bng
 
+# Python 2
```
jmuhlich: Same.

johnbachman:
Updated the comments as suggested.

PR #236, opened 2016-11-02
Logging framework & Simulator class multiple initials/params support

alubbock:
This PR contains two main features:

* A new logging framework for PySB, built on the Python logging module, just integrated into the Simulation class for now, but which will facilitate new capabilities:
  * Finer-grained verbosity selection than verbose=True or False using logging levels.
  * Support for multiple logging handlers, allowing logging to console, file, various centralized network servers. This will be useful for distributed/parallel/cluster-based PySB setups.
  * We've discussed a benchmark suite before. A custom benchmarking log handler could be written to time various pieces of code based on start/stop events.
* The ability to run simulations using multiple different initial condition/parameter value sets. This code was written by @lh64. This capability is also important for the upcoming cupSODA simulator, which can run multiple such simulations in parallel.

I also removed the extraneous output from `weave_inline` when running simulations by capturing STDERR and redirecting it to /dev/null (in commit 9471277).

### DIFF
--- a/pysb/examples/michment.py
+++ b/pysb/examples/michment.py
@@ -1,0 +1,26 @@
+from pysb import *
+from pysb.macros import catalyze_state
+
+Model()
+
+VOL = 1. # volume (arbitrary units)
+
+Parameter('kf',   1./VOL)
+Parameter('kr',   1000)
+Parameter('kcat', 100)
+
+Monomer('E', ['s'])
+Monomer('S', ['e', 'state'], {'state': ['0','1']})
+
+catalyze_state(E(), 's', S(), 'e', 'state', '0', '1', [kf,kr,kcat])
+
+Observable("E_free",     E(s=None))
+Observable("S_free",     S(e=None, state='0'))
+Observable("ES_complex", E(s=1) % S(e=1))
+Observable("Product",    S(e=None, state='1'))
+
+Parameter("Etot", 1.*VOL)
+Initial(E(s=None), Etot)
+
+Parameter('S0', 10.*VOL)
+Initial(S(e=None, state='0'), S0)
--- a/pysb/examples/michment.py
+++ b/pysb/examples/michment.py
@@ -3,9 +3,9 @@ from pysb.macros import catalyze_state
 
 Model()
 
-VOL = 1. # volume (arbitrary units)
+Parameter('vol', 1.) # volume (arbitrary units)
 
-Parameter('kf',   1./VOL)
+Parameter('kf',   1./vol.value)
 Parameter('kr',   1000)
 Parameter('kcat', 100)
 
@@ -19,8 +19,8 @@ Observable("S_free",     S(e=None, state='0'))
 Observable("ES_complex", E(s=1) % S(e=1))
 Observable("Product",    S(e=None, state='1'))
 
-Parameter("Etot", 1.*VOL)
+Parameter("Etot", 1.*vol.value)
 Initial(E(s=None), Etot)
 
-Parameter('S0', 10.*VOL)
+Parameter('S0', 10.*vol.value)
 Initial(S(e=None, state='0'), S0)
--- a/pysb/integrate.py
+++ b/pysb/integrate.py
@@ -41,9 +41,15 @@ class Solver(object):
     yobs : numpy.ndarray with record-style data-type
         Observable trajectories. Length is ``len(tspan)`` and record names
         follow ``model.observables`` names.
+    yobs_view : numpy.ndarray
+        An array view (sharing the same data buffer) on ``yobs``.
+        Dimensionality is ``(len(tspan), len(model.observables))``.
     yexpr : numpy.ndarray with record-style data-type
         Expression trajectories. Length is ``len(tspan)`` and record names
         follow ``model.expressions_dynamic()`` names.
+    yexpr_view : numpy.ndarray
+        An array view (sharing the same data buffer) on ``yexpr``.
+        Dimensionality is ``(len(tspan), len(model.expressions_dynamic()))``.
     integrator : scipy.integrate.ode
         Integrator object.
 
@@ -64,6 +70,8 @@ class Solver(object):
                                      integrator=integrator, cleanup=cleanup,
                                      **integrator_options)
         self.result = None
+        self._yexpr_view = None
+        self._yobs_view = None
 
     @property
     def _use_inline(self):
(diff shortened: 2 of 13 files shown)


jmuhlich:
@lh64 One can still have network-oriented simulations with no initial conditions, in the presence of synthesis reactions. So pre-filling model.species is not a solution. It may be good enough to write the check as a non-public method on the base Simulator class, and document that any subclasses with enumerated species should call that method in their setup process. Or we can try mixins or an inheritance-based solution.

pysb/simulator/base.py
```diff
@@ -108,29 +116,61 @@ def initials(self, new_initials):
                                                  'MonomerPattern or '
                                                  'ComplexPattern' %
                                                  repr(cplx_pat))
+                    # if val is a number, convert it to a single-element array
+                    if not isinstance(val, collections.Iterable):
```
jmuhlich: I think this should be `Sequence` instead of `Iterable`. You need `Sequence` if you want to retrieve an element from a specific index, which I believe the code using these values will do. I see this mistake in a bunch of other places. I'll try to make a note on the others.

pysb/simulator/base.py
```diff
@@ -139,20 +179,44 @@ def _set_initials(initials_source):
...
-
+                # Loop over all simulations
+                for sim in range(len(y0)):
+                    # If this initial condition has already been set, skip it
+                    # (i.e., an override)
+                    if y0[sim][si] != 0:
+                        continue
+                    
+                    def _get_value(sim):
+                        if isinstance(value_obj, collections.Iterable) and \
```
jmuhlich: Should be `Sequence` not `Iterable`.

pysb/simulator/base.py
```diff
@@ -186,27 +266,92 @@ def param_values(self, new_params):
...
         if isinstance(new_params, dict):
-            for k in new_params.keys():
-                if k not in self._model.parameters.keys():
+            for key,val in new_params.items():
+                if key not in self._model.parameters.keys():
                     raise IndexError("new_params dictionary has unknown "
-                                     "parameter name (%s)" % k)
+                                     "parameter name (%s)" % key)
+                # if val is a number, convert it to a single-element array
+                if not isinstance(val, collections.Iterable):
```
jmuhlich: Should be `Sequence` not `Iterable`.

pysb/simulator/base.py
```diff
@@ -322,12 +474,12 @@ def __init__(self, simulator, trajectories):
                 raise ValueError("trajectories should be a 3D array or a list "
                                  "of 2D arrays")
             self._y = trajectories
-
-        self.nsims = len(self._y)
-        if len(self.tout) != self.nsims:
+        
+        self._nsims = len(self._y)
```
jmuhlich: I think it could be useful to keep this public, but read-only. Can we have a getter `nsims` that returns `self._nsims`? It would clean up a lot of the simulator code too, to revert all uses of `self._nsims` to `self.nsims`.

pysb/logging.py
```diff
@@ -0,0 +1,146 @@
...
+    log_fmt = logging.Formatter('%(asctime)s.%(msecs).3d - %(name)s - '
+                                '%(levelname)s - %(message)s',
+                                datefmt='%Y-%m-%d %H:%M:%S')
+    if time_utc:
+        log_fmt.converter = time.gmtime
+
+    return log_fmt
+
+
+def setup_logger(level=logging.INFO, console_output=True, file_output=False,
```
jmuhlich: Default should be WARNING. I think INFO is too verbose for the default.

pysb/logging.py
```diff
@@ -0,0 +1,146 @@
...
+    Returns
+    -------
+    A logging.Logger object for PySB logging. Note that other PySB modules
+    should use a logger specific to their namespace instead by calling
+    :func:`get_logger`.
+    """
+    log = logging.getLogger(BASE_LOGGER_NAME)
+
+    # Logging level can be overridden with environment variable
+    if DEBUG_ENV_VAR in os.environ:
```
jmuhlich: The env var should allow specific levels, not just DEBUG, especially if the default is changed to WARNING (might be nice to get INFO level sometimes but not full DEBUG). Possibly the var name could change to PYSB_VERBOSE and it could take integer levels or string names (same as the level names from `logging`).

pysb/simulator/base.py
```diff
@@ -73,9 +75,17 @@ class Simulator(object):
...
     __metaclass__ = ABCMeta
 
+    _supports = { 'multi_initials' : False,
+                  'multi_param_values' : False }
+
     @abstractmethod
     def __init__(self, model, tspan=None, initials=None,
                  param_values=None, verbose=False, **kwargs):
+        # Get or create base PySB logger exists
+        self._logger = get_logger(self.__module__)
```
jmuhlich: You should use a custom `logging.LoggerAdapter` subclass here that takes `model.name` as an extra constructor arg, and implements the `process` method, returning `('[%s] %s' % (self.name, msg), kwargs)`. Then all calls to that logger will automatically prepend `[model_name]` instead of having to remember to do that explicitly at all call sites!

pysb/simulator/scipyode.py
```diff
@@ -299,78 +302,82 @@ def jacobian(t, y, p):
...
+
+        Produces compile warnings, which we suppress by capturing STDERR.
+        """
         if not hasattr(cls, '_use_inline'):
             cls._use_inline = False
             try:
                 if weave_inline is not None:
-                    weave_inline('int i=0; i=i;', force=1)
+                    extra_compile_args = None
+                    if os.name == 'posix':
```
jmuhlich: Nice work on this.

pysb/examples/michment.py
```diff
@@ -0,0 +1,26 @@
+from pysb import *
```
lh64: It encodes the Michaelis-Menten reaction set: E + S <-> ES -> E + P

pysb/integrate.py
```diff
@@ -41,15 +41,9 @@ class Solver(object):
     yobs : numpy.ndarray with record-style data-type
         Observable trajectories. Length is ``len(tspan)`` and record names
         follow ``model.observables`` names.
-    yobs_view : numpy.ndarray
```
alubbock: Fixed in 3f59f27

alubbock:
@jmuhlich I've made changes incorporating all your feedback -- thanks.

pysb/examples/michment.py
jmuhlich: I think when people hear "Michaelis-Menten" they generally think of the M-M equation for the reaction velocity rather than that piecewise mass-action decomposition. Would it be OK if I later revise this to use the actual M-M equation, and create a parallel model that uses the mass-action formulation? We could then have a run script that simulates both over some parameter ranges and plots the differences.

jmuhlich:
This looks good. Will merge today.

pysb/examples/michment.py
lh64: Sure, that's fine. Maybe just call this one michment_massaction.py or something.

PR #237, opened 2017-01-11
Fix SBML export species id mismatch with leading underscores

jmuhlich:
The SBML exporter makes some assumptions about how PySB names species,
and this changed in the BNG interface code which creates the species
but wasn't updated to match here.


bgyori:
Looks good, thanks!

PR #242, opened 2017-02-15
cupSODA simulator support

alubbock:
This PR adds PySB support for the cupSODA simulator, which is an implementation of LSODA for NVIDIA GPUs using CUDA. More info and cupSODA downloads are at https://github.com/aresio/cupSODA (the easiest way to get started is to download a binary from [releases](https://github.com/aresio/cupSODA/releases)).

GPU-based unit tests are skipped on Travis due to a lack of GPU support, but the unit test suite completes successfully on our test Win/Mac/Linux machines.

cupSODA currently only supports mass action kinetic models. We're looking into adding support for other rate law types in the future.


pysb/simulator/base.py
```diff
@@ -400,22 +411,22 @@ class SimulationResult(object):
...
-       2.2344e-10   2.6394e-12   4.8067e-14  -6.2097e-14  -7.5308e-15]
+    [  1.0000e+00   1.1744e-02   1.3791e-04   1.6196e-06   1.9020e-08
+       2.2337e-10   2.6232e-12   3.0806e-14   3.6178e-16   4.2492e-18]
 
     It is also possible to retrieve the value of all observables at a
     particular time point, e.g. the final concentrations:
 
-    >>> print(simulation_result.observables[-1]) #doctest: +ELLIPSIS
-    (-7.5308...e-15, -1.6809...e-13, 1.0000...)
+    >>> obs_last_timepoint = simulation_result.observables[-1]
```
jmuhlich: Hah, we just discovered this issue on our side too. I think it's specifically numpy 1.12 fixed a bug with set_printoptions(precision=0) but in the process changed how `numpy.void` objects are formatted: trailing zeros after the decimal place are suppressed. So 1.0000 becomes "1." and 1.2300 becomes "1.23". Rows from a recarray are returned as `void` which is why we only have this problem when displaying a row from `observables` or `all`. (Presumably `expressions` would have the same problem). I would suggest we go back to the old way of printing the numbers in the doctest, but format the doctest output to match the numpy 1.12 way. Since our Travis setup will use the latest numpy, CI will pass. Anyone running doctests locally will just need the latest numpy (but I don't think we should require '>=12' in `setup.py`.

pysb/simulator/base.py
```diff
@@ -430,19 +441,18 @@ class SimulationResult(object):
...
+     [  3.6178e-16   1.2863e-14   1.0000e+00]
+     [  4.2492e-18   1.6996e-16   1.0000e+00]]
 
     Species, observables and expressions can be combined into a single numpy
     ndarray and accessed similarly. Here, the initial concentrations of all
     these entities are examined:
 
-    >>> print(simulation_result.all[0])
-    (1.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0)
+    >>> init_concs = simulation_result.all[0]
```
jmuhlich: Same as above - suggest we go back to printing out but remove the trailing zeros.

pysb/simulator/cupsoda.py
```diff
@@ -0,0 +1,595 @@
...
+            raise Exception('memory_usage must be one of %s',
+                            self._memory_options.keys())
+
+    @property
+    def vol(self):
+        vol = self.opts.get('vol', None)
+        return vol
+
+    @vol.setter
+    def vol(self, volumne):
```
jmuhlich: Harmless typo. :)

pysb/simulator/cupsoda.py
```diff
@@ -0,0 +1,595 @@
...
+        self._len_rxns = len(self._model.reactions)
+        self._len_species = len(self._model.species)
+        self._len_params = len(self._model.parameters)
+        self._model_parameters_rules = self._model.parameters_rules()
+
+        # Set cupsoda verbosity level
+        logger_level = self._logger.logger.getEffectiveLevel()
+        if logger_level > logging.DEBUG:
+            self._cupsoda_verbose = 0
+        elif logger_level <= 5:
```
jmuhlich: I think it might be more readable to reorganize this so the tested ranges progress in order rather than skipping around. Also to document what level 5 is via a named constant (no magic numbers). Something like this:

MAX_DEBUG = 5 (at the top level)
...
if level <= MAX_DEBUG ... elif level <= logging.DEBUG ... else ...

alubbock:
@jmuhlich Thanks for the review. All feedback implemented. I added an `EXTENDED_DEBUG` log level (value=5) to logging.py, so it can be used by other parts of PySB, not just cupSODA. Unit tests pass in both Python 2.7 and 3.5.

pysb/simulator/cupsoda.py
```diff
@@ -0,0 +1,596 @@
...
+
+        # private variables (to reduce the number of function calls)
+        self._len_rxns = len(self._model.reactions)
+        self._len_species = len(self._model.species)
+        self._len_params = len(self._model.parameters)
+        self._model_parameters_rules = self._model.parameters_rules()
+
+        # Set cupsoda verbosity level
+        logger_level = self._logger.logger.getEffectiveLevel()
+        if logger_level > logging.DEBUG:
```
jmuhlich: My other suggestion here was to reorder the if/else clauses in a more logical order, probably from most to least verbose (see my previous comment for proposed code). If there's a reason to keep it in this order that's OK too, I just found I had to read the code more closely than I'd like to understand the intent.

jmuhlich: Here's the pseudocode from my earlier comment:
``` if level <= EXTENDED_DEBUG ... elif level <= logging.DEBUG ... else ...```

Actually that brings up another point -- could you make the new level (5) available directly in a top level symbol, e.g. `pysb.logging.EXTENDED_DEBUG` ? Would be easier to use then the dict lookup.

jmuhlich:
Beautiful, thanks!

PR #246, opened 2017-03-25
Initial concentration sensitivity analysis tool

alubbock:
A new tool for sensitivity analysis by varying pairwise combinations of species initial concentrations and examining the response of a model output/observable. Includes functions for visualising the results with matplotlib and a worked example using the 1991 Tyson cell cycle model.

### DIFF
--- a/pysb/tools/sensitivity_analysis.py
+++ b/pysb/tools/sensitivity_analysis.py
@@ -13,6 +13,7 @@ from matplotlib import gridspec, pyplot
 import numpy as np
 import pysb.integrate
 from pysb.bng import generate_equations
+matplotlib.rcParams['svg.fonttype'] = 'none'
 matplotlib.use('Agg')
 plt = matplotlib.pyplot
 
@@ -47,10 +48,8 @@ class InitialConcentrationSensitivityAnalysis:
         self.objective_function = objective_function
         self.observable = observable
         self.standard = None
-        self.sensitivity_matrix = None
-        self.sensitivity_list = None
-        self.p_prime_matrix = np.array([])
-        self.p_matrix = np.array([])
+        self.p_prime_matrix = np.zeros(self.size_of_matrix)
+        self.p_matrix = np.zeros(self.size_of_matrix)
 
     def _calculate_objective(self, species):
         """ Calculates fraction of change between standard value and newly obtained value
@@ -132,17 +131,14 @@ class InitialConcentrationSensitivityAnalysis:
 
         a_matrix = cartesian_product(self.values_to_sample,
                                      self.index_of_species_of_interest)
-        a_matrix = a_matrix.T.reshape(
-                np.shape(a_matrix)[0] * np.shape(a_matrix)[1], )
+        a_matrix = a_matrix.T.reshape(self.n_sam * self.n_proteins)
+
         a_prime = cartesian_product(np.ones(self.n_sam),
                                     self.index_of_species_of_interest)
-        a_prime = a_prime.T.reshape(
-                np.shape(a_prime)[0] * np.shape(a_prime)[1], )
-        self.b_matrix = cartesian_product(a_matrix, a_matrix)
+        a_prime = a_prime.T.reshape(self.n_sam * self.n_proteins)
 
+        self.b_matrix = cartesian_product(a_matrix, a_matrix)
         self.b_prime_matrix = cartesian_product(a_prime, a_matrix)
-        np.set_printoptions(linewidth=12000, threshold=np.inf)
-        # print(self.b_prime_matrix)
 
     def _g_function(self, matrix):
         """
@@ -153,7 +149,8 @@ class InitialConcentrationSensitivityAnalysis:
         counter = -1
         sampled_values_index = set()
         bij_unique = dict()
-        sampling_matrix = np.zeros((self.size_of_matrix, len(self.model.species)))
+        sampling_matrix = np.zeros(
+                (self.size_of_matrix, len(self.model.species)))
         sampling_matrix[:, :] = self.initial_conditions
         for j in range(len(matrix)):
             for i in matrix[j, :]:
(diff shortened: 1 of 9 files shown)


pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,603 @@
...
+        save_name : str, optional
+            prefix of saved files
+        out_dir : str, optional
+            location to save output if required
+        """
+        if out_dir is not None:
+            if not os.path.exists(out_dir):
+                os.mkdir(out_dir)
+        if self._solver is None:
+            raise(TypeError, "Must provide a pysb.simulator to run "
```
jmuhlich: How does one provide a Simulator to the run method?

pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,603 @@
...
+        self.p_prime_matrix = p_prime_matrix.reshape(self._shape_of_matrix)
+
+        # save output if desired
+        if save_name is not None:
+            if out_dir is None:
+                out_dir = '.'
+            p_name = os.path.join(out_dir, '{}_p_matrix.csv'.format(save_name))
+            p_prime_name = os.path.join(
+                out_dir, '{}_p_prime_matrix.csv'.format(save_name))
+            print("Saving p matrix and p' matrix to {} and {}".format(
```
jmuhlich: Should use logging system instead of print. There are a couple of other instances of this in here too.

pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,603 @@
...
+
+        self._original_initial_conditions = np.zeros(len(self._model.species))
+        self._index_of_species_of_interest = self._create_index_of_species()
+        self.simulation_initials = self._setup_simulations()
+        # Stores the objective function value for the original unperturbed
+        # model
+        self._objective_fn_standard = None
+
+    @property
+    def _proteins_of_interest(self):
```
jmuhlich: Should this be called `parameters_of_interest`?

pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,603 @@
...
+            print("Saving p matrix and p' matrix to {} and {}".format(
+                p_name, p_prime_name))
+
+            np.savetxt(p_name, self.p_matrix)
+            np.savetxt(p_prime_name, self.p_prime_matrix)
+
+    def _create_index_of_species(self):
+        """ create dictionary of initial conditions by index """
+        index_of_init_condition = {}
+        for i in range(len(self._model.initial_conditions)):
```
jmuhlich: Iterating over `initial_conditions` and `species` directly with good variable names would make this dense code a lot more readable. I see you do need `j`, so `enumerate` could be used there. Also note that since `initial_conditions` entries are tuples, you can unpack both entries into named variables.

pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,603 @@
...
+            for j in range(len(self._model.species)):
+                if str(self._model.initial_conditions[i][0]) \
```
jmuhlich: `ComplexPattern` has an `is_equivalent_to` method that should be used here rather than comparing the results of `str`.

pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,603 @@
...
+    array_1 : list_like
+    array_2 : list_like
+
+    Returns
+    -------
+    out : np.array
+        array of shape (len(array_1) x len(array_2))
+    """
+    a = list(product(array_1, array_2))
+    a = np.asarray(a, dtype=','.join('object' for _ in range(len(a[0]))))
```
jmuhlich: A more concise way to write this is `.join(['object'] * len(a[0]))`.

pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,603 @@
...
+            if out_dir is None:
+                out_dir = '.'
+            if not os.path.exists(out_dir):
+                os.mkdir(out_dir)
+            fig.savefig(os.path.join(out_dir,
+                                     '{}_P_H_P_prime.png'.format(save_name)),
+                        bbox_inches='tight')
+        if show:
+            plt.show()
+        plt.close()
```
jmuhlich: Could return `fig` here just in case someone wants to do something with it you didn't anticipate. Same goes for the other plotting functions below.

glucksfall:
Hello everyone,
I'm interested in this feature. It's helpful to calculate how much sensitives are different parameters (either if they are rates, initials, whatever...). However, I just wonder how do you do this. You use an 'equivalent' ODE model to calculate sensitivity, or there is a direct method to calculate sensitivity from a stochastic simulation? Also, why the analysis is pairwise?

alubbock:
@glucksfall This PR only implements sensitivity to changes in initials, but we are interested in adding support for rate parameters in the future. It takes a simulator from a `pysb.simulator.Simulator` subclass and calculates sensitivities empirically by running simulations on it. The analysis is a good use case for our recently added `CupSodaSimulator`, an ODE simulator for NVIDIA GPUs, which is ideally suited for this task due to its parallel capabilities.

Currently, all of PySB's Simulator classes are ODE-based, but we will be adding stochastic ones soon (an SSA implementation is top of the list).

The analysis is pairwise because it calculates changes in sensitivity (as measured by a PySB `Observable`) to changes in two initial conditions at a time, although sensitivity to changes in single initial conditions can also be shown.

pysb/tools/sensitivity_analysis.py
```diff
@@ -0,0 +1,627 @@
...
+
+        self._original_initial_conditions = np.zeros(len(self._model.species))
+        self._index_of_species_of_interest = self._create_index_of_species()
+        self.simulation_initials = self._setup_simulations()
+        # Stores the objective function value for the original unperturbed
+        # model
+        self._objective_fn_standard = None
+
+    @property
+    def _species_of_interest(self):
```
jmuhlich: Species is definitely more appropriate than proteins.. :)  But my point was that this appears to store **parameters** not **species**. Initial condition parameters, yes, but it's still a misleading name. Maybe `_ic_params_of_interest`?

PR #247, opened 2017-03-28
Import from BioModels; better SBML expression handling

alubbock:
A small commit which improves the BNGL and SBML importer expression handling, allowing parameters, observables and other expressions as symbols within an expression.

I've also added a convenience function `model_from_biomodels` which downloads an SBML model from [BioModels](https://www.ebi.ac.uk/biomodels-main/) and runs it through the SBML translator in a one-liner (see example doctest).

### DIFF
--- a/pysb/importers/sbml.py
+++ b/pysb/importers/sbml.py
@@ -12,6 +12,7 @@ try:
 except ImportError:
     # Python 2
     from urllib import urlretrieve
+from pysb.logging import get_logger, EXTENDED_DEBUG
 
 BIOMODELS_REGEX = re.compile(r'BIOMD[0-9]{10}')
 BIOMODELS_URL = 'http://www.ebi.ac.uk/biomodels-main/download?mid={}'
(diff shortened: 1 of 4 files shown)


pysb/importers/sbml.py
```diff
@@ -153,7 +163,69 @@ def model_from_sbml(filename, force=False, cleanup=True, **kwargs):
...
+    >>> print(model) #doctest: +ELLIPSIS
+    <Model 'pysb' (monomers: 12, rules: 17, parameters: 37, expressions: 0, ...
+    """
+    if not BIOMODELS_REGEX.match(accession_no):
+        try:
+            accession_no = 'BIOMD{:010d}'.format(int(accession_no))
+        except ValueError:
+            raise ValueError('accession_no must be an integer or a BioModels '
+                             'accession number (BIOMDxxxxxxxxxx)')
+    print('Retrieving {}'.format(accession_no))
```
jmuhlich: Should use logging.

PR #249, opened 2017-04-05
Add StochKit interface

alubbock:
Add interface for [StochKit](https://github.com/stochss/stochkit), a stochastic simulation toolkit with implementations of the Gillespie stochastic simulation algorithm (SSA) and tau-leaping algorithm. It contains an exporter for StochKit's file format and a `Simulation` class interface.


pysb/simulator/base.py
```diff
@@ -201,7 +201,9 @@ def _get_value(sim):
                         elif isinstance(value_obj, Component):
                             if value_obj in self._model.parameters:
                                 pi = self._model.parameters.index(value_obj)
-                                value = self.param_values[sim][pi]
+                                value = self.param_values[
+                                    sim if len(self.param_values) > 1 else
```
jmuhlich: This is awkward wrapped over so many lines. Suggest this:
```python
pset_idx = sim if len(self.param_values) > 1 else 0
value = self.param_values[pset_idx][pi]
```

pysb/simulator/stochkit.py
```diff
@@ -0,0 +1,305 @@
...
+        if epsilon is not None:
+            extra_args += ' --epsilon {:f}'.format(epsilon)
+
+        if threshold is not None:
+            extra_args += ' --threshold {:d}'.format(threshold)
+
+        # Find binary for selected algorithm (SSA, Tau-leaping, ...)
+        if algorithm is None:
+            raise SimulatorException("No StochKit algorithm selected")
+        if not re.match(r'\w', algorithm):
```
jmuhlich: This just checks that there is *at least one* letter or digit in `algorithm`. You'd need to switch the sense of the test and/or modify the regex to achieve what was intended based on the exception message. Regardless, the `not in` check below is an even stronger trap for bad values, so this `re.match` test can be dropped entirely.

pysb/simulator/stochkit.py
```diff
@@ -0,0 +1,305 @@
...
+            extra_args += ' --method {}'.format(method)
+
+        if epsilon is not None:
+            extra_args += ' --epsilon {:f}'.format(epsilon)
+
+        if threshold is not None:
+            extra_args += ' --threshold {:d}'.format(threshold)
+
+        # Find binary for selected algorithm (SSA, Tau-leaping, ...)
+        if algorithm is None:
```
jmuhlich: If the default was `None` then this would be a good error message to have. However the default is 'ssa' so someone would have to explicitly pass `None` to fail this check, and there's a test against an explicit list of valid simulator names below anyway, so you can drop this.

pysb/simulator/stochkit.py
```diff
@@ -0,0 +1,305 @@
...
+                '{0}' output:\nSTDOUT:\n{1}\nSTDERR:\n{2}".format(
+                    cmd, stdout, stderr))
+
+            traj_dir = os.path.join(prefix_outdir, 'trajectories')
+            try:
+                trajectories.extend([np.loadtxt(os.path.join(
+                    traj_dir, f)) for f in sorted(os.listdir(traj_dir))])
+            except Exception as e:
+                raise SimulatorException(
+                    "Error using solver.get_trajectories('{0}'): {1}".format(
```
jmuhlich: I think this error message needs some tweaking -- I can't figure out what `get_trajectories` is referring to.

jmuhlich:
I think the merge/rebase on this one might be rough...

PR #250, opened 2017-04-05
Read Expression values from BNG .gdat files

lh64:
Modified pysb.bng.BngBaseInterface.read_simulation_results() to read in header names from the .gdat file rather than set the names equal to the names of model.observables. This is necessary because BNG can be instructed to print function (Expression) values to the .gdat file with the 'print_functions=1' argument. This also required changing the 'if' statement from a check on whether model.observables exists to whether the .gdat file exists (since, in principle, the .gdat file could contain only functions).

### DIFF
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -182,26 +182,26 @@ class BngBaseInterface(object):
         """
         # Read concentrations data
         cdat_arr = numpy.loadtxt(self.base_filename + '.cdat', skiprows=1)
+        names = ['time'] + ['__s%d' % i for i in range(cdat_arr.shape[1]-1)] # -1 for time column
+
         # Read groups data
-        if self.model and len(self.model.observables):
+        if os.path.isfile(self.base_filename + '.gdat'):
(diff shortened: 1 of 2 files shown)


pysb/bng.py
```diff
@@ -179,28 +180,49 @@ def read_simulation_results(self):
...
-            gdat_arr = numpy.loadtxt(self.base_filename + '.gdat',
-                                     skiprows=1)[:,1:]
-        else:
+        try:
+            with open(self.base_filename + '.gdat', 'r') as f:
+                # Exclude \# and time column
+                names += f.readline().split()[2:]
+                # Exclude first column (time)
+                gdat_arr = numpy.loadtxt(f)
+                print(gdat_arr)
```
jmuhlich: Remove debugging print.

pysb/bng.py
```diff
@@ -14,6 +14,7 @@
 import shutil
 import collections
 import pysb.pathfinder as pf
+from pysb.simulator import SimulationResult
```
jmuhlich: Remove unused import.

Issue #210, opened 2016-05-18
PySB Flat exporter fails when a Compartment is defined with a size Parameter inline


alubbock:
By defining a size Parameter inline in the Compartment declaration, the Parameter() definition gets repeated twice (once in the Parameters list, once in the Compartment list), leading to a `ComponentDuplicateNameError`.

Minimal example:

``` python
from pysb import *
from pysb.export.pysb_flat import PysbFlatExporter
Model()
Compartment('C1', None, 3, Parameter('C1_size', 1.0))
print(PysbFlatExporter(model).export())
```

Output Python code:

``` python
# exported from PySB model '_interactive_'

from pysb import Model, Monomer, Parameter, Expression, Compartment, Rule, Observable, Initial, Annotation, ANY, WILD

Model()

Parameter('C1_size', 1.0)

Compartment(name='C1', parent=None, dimension=3, size=Parameter('C1_size', 1.0))
```

Error when running the output Python code:

```
ComponentDuplicateNameError: Tried to add a component with a duplicate name: C1_size
```

PR #254, opened 2017-04-19
Fix PysbFlatExporter with compartmental models

alubbock:
The PysbFlatExporter fails with compartmental models due to `Compartment`'s
`repr()` implementation, which means parameters and parent compartments
get printed in full rather than just their name, which leads to a
`ComponentDuplicateNameError` with compartmental models when using the
PysbFlatExporter. This PR fixes the issue.

Resolves: pysb/pysb#210

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -857,8 +857,13 @@ class Compartment(Component):
         self.size = size
 
     def __repr__(self):
-        return  '%s(name=%s, parent=%s, dimension=%s, size=%s)' % \
-            (self.__class__.__name__, repr(self.name), repr(self.parent), repr(self.dimension), repr(self.size))
+        return '%s(name=%s, parent=%s, dimension=%s, size=%s)' % (
+            self.__class__.__name__,
+            repr(self.name),
+            'None' if self.parent is None else self.parent.name,
+            repr(self.dimension),
+            'None' if self.size is None else self.size.name
+        )
 
 
 class Rule(Component):
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1468,10 +1468,10 @@ class Model(object):
         source_cp = as_complex_pattern(self.monomers['__source']())
         if self.compartments:
             for c in self.compartments:
-                source_cp.compartment = c
-                if not any(source_cp.is_equivalent_to(other_cp) for
+                source_cp_cpt = source_cp ** c
+                if not any(source_cp_cpt.is_equivalent_to(other_cp) for
                            other_cp, value in self.initial_conditions):
-                    self.initial(source_cp, self.parameters['__source_0'])
+                    self.initial(source_cp_cpt, self.parameters['__source_0'])
         else:
             if not any(source_cp.is_equivalent_to(other_cp) for other_cp, value in self.initial_conditions):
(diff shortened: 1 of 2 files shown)


jmuhlich:
@alubbock needs a `rebase -i` against master, dropping a85b119.

alubbock:
@jmuhlich Done.

pysb/core.py
```diff
@@ -516,15 +518,22 @@ def is_equivalent_to(self, other):
         """Checks for equality with another ComplexPattern"""
         # Didn't implement __eq__ to avoid confusion with __ne__ operator used for Rule building
 
+        # Check both patterns are concrete
```
jmuhlich: I did just verify that pysb itself does only use this to compare concrete species CPs so this won't break any internals, but it *is* a breaking change to the public API -- perhaps someone else was depending on being able to compare non-concrete CPs in their model or framework. Therefore this should merely be a DeprecationWarning.

The docstring needs to change to mention the concreteness requirement, while mentioning that passing non-concrete CPs is deprecated. Suggested new docstring:

```python
"""
Test a concrete ComplexPattern for equality with another.

Use of this method on non-concrete ComplexPatterns was previously allowed, but is now deprecated.

"""

File: pysb/core.py [hunk 2]
```diff
@@ -1444,7 +1460,7 @@ def enable_synth_deg(self):
         source_cp = as_complex_pattern(self.monomers['__source']())
         if self.compartments:
             for c in self.compartments:
-                source_cp = source_cp ** c
+                source_cp.compartment = c
```
jmuhlich: Just avoiding usage of the overloaded operators within the core? I thought that might be a good style policy but it's not written down anywhere.

alubbock: The change is needed to avoid the new `CompartmentAlreadySpecifiedError`

jmuhlich: OK. One issue I notice here is that each time through the loop, the same CP object is being modified in place then used as the IC pattern. The code path for adding an IC doesn't copy the pattern, so I would think the ICs will actually all be added using the same pattern. This should trigger an exception in `initial` so I'm not confident I'm reading this right. When I get back to my desk I'll investigate further.

alubbock: You're right, there's an error. `source_cp` gets added to initials assigned to the first compartment in the `for` loop, then its compartment gets updated **in place**, such that `is_equivalent_to` returns true on subsequent loop iterations (since the patterns are the same object) and the pattern doesn't get added again. The end result is that we only have one `source()` which is assigned to the last compartment in the `for` loop due to the in place compartment update.

I'll write a patch using `copy()` and push it here.

pysb/core.py
```diff
@@ -1468,10 +1468,10 @@ def enable_synth_deg(self):
         source_cp = as_complex_pattern(self.monomers['__source']())
         if self.compartments:
             for c in self.compartments:
-                source_cp.compartment = c
-                if not any(source_cp.is_equivalent_to(other_cp) for
+                source_cp_cpt = source_cp ** c
```
jmuhlich: This is fine, but the original code was OK too as `**` does return a copy! :)

jmuhlich: Oops nevermind, I misunderstood the problem yet again.

PR #255, opened 2017-04-19
Add python3 support to cupSODA simulator

alubbock:
The current unit test suite for the cupSODA interface fails on Python 3.
This commit has a few small changes to add Python 3 compatibility and
fix the unit test suite.


alubbock:
Travis is currently failing due to an SSL error which was resolved in another PR (pysb/pysb#254). When that one is merged, a rebase should fix the Travis error here. Unit tests pass locally.

Issue #259, opened 2017-04-21
Reaction rate params/expressions matching sympy keywords cause an error with `generate_equations`


alubbock:
Defining a parameter or expression using a sympy function name (e.g. `deg`) will cause the following error:

    SympifyError: SympifyError: <function deg at 0x10f0d0140>

The error is down to `_parse_reaction` in bng.py, which runs `sympy.S(x)` where `x` is a string such as `deg`. Converting the string to the relevant parameter or expression object should resolve this.

PR #260, opened 2017-04-21
Allow rxn rate params/exprs with sympy fn names

alubbock:
Defining a parameter or expression using a sympy function name (e.g.
`deg`) will cause the following error:

    SympifyError: SympifyError: <function deg at 0x10f0d0140>

The error is down to `_parse_reaction` in bng.py, which runs `sympy.S(x)`
where `x` is a string such as `'deg'`. Converting the string to the relevant
parameter or expression object resolves this.

Resolves: pysb/pysb#259

### DIFF
--- a/pysb/simulator/cupsoda.py
+++ b/pysb/simulator/cupsoda.py
@@ -1,5 +1,6 @@
 from __future__ import print_function as _
 from pysb.simulator.base import Simulator, SimulatorException, SimulationResult
+import pysb
 import pysb.bng
 import numpy as np
 from scipy.constants import N_A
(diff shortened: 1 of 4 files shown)


pysb/simulator/cupsoda.py
```diff
@@ -474,12 +474,13 @@ def _get_cmatrix(self):
             for j in range(self._len_rxns):
                 rate = 1.0
                 for r in rate_args[j]:
-                    x = str(r)
-                    if x in par_names:
-                        rate *= par_vals[i][par_dict[x]]
+                    if isinstance(r, pysb.Parameter):
+                        rate *= par_vals[i][par_names.index(r.name)]
```
jmuhlich: Was going through `par_dict` causing a problem, or did you decide it was just an unnecessary optimization?

alubbock: It didn't cause a problem, but I was simplifying the code when making some other changes and didn't revert this change back. I would imagine the performance difference is negligible in any case.

PR #263, opened 2017-05-01
Continuous integration on Windows (appveyor)

alubbock:
This PR adds continuous integration on Windows using the [Appveyor](https://ci.appveyor.com/) platform. Enable using the Appveyor Github integration.

Notes:

- I've set up tests for Python 2.7 and Python 3.4 on win32 platform
- PyGraphviz binaries from [Chris Gohlke](http://www.lfd.uci.edu/~gohlke/pythonlibs/) are used, but can't be downloaded directly, so I used Dropbox. The files are <1Mb so performance shouldn't be impacted.
- Kappa doesn't have a `KaSa` binary available for Windows AFAIK, so the Kappa static analysis tests are skipped on that platform (Kappa simulations are still tested)
- Anaconda doesn't have a numpy 1.12 binary for Python 3.4/win32. I added doctest skip declarations to two tests in simulator/base.by to make the test suite pass on numpy 1.11.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -45,6 +45,11 @@ install:
   - appveyor DownloadFile "https://github.com/Kappa-Dev/KaSim/releases/download/v4beta/KaSa.exe" -FileName KaSa.exe
   - set "KAPPAPATH=%cd%"
 
+  # StochKit
+  - appveyor DownloadFile "https://downloads.sourceforge.net/project/stochkit/StochKit2/StochKit2.0.10/StochKit2.0.10_WINDOWS_LITE.zip" -FileName stochkit.zip
+  - 7z x stochkit.zip -y > nul
+  - set "STOCHKITPATH=%cd%\StochKit2.0.10_WINDOWS_LITE"
+
   # Build PySB
   - python setup.py build --build-lib=build/lib
 test_script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,50 @@
+build: false
+
+environment:
+  matrix:
+    - PYTHON_VERSION: 2.7
+      PYTHON_SVER: 27
+      MINICONDA: C:\Miniconda
+    - PYTHON_VERSION: 3.4
+      PYTHON_SVER: 34
+      MINICONDA: C:\Miniconda3
+
+init:
+  - "ECHO %PYTHON_VERSION% %MINICONDA%"
+
+install:
+  # Issues have been encountered with installing numpy and scipy on
+  # AppVeyor e.g.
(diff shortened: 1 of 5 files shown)


appveyor.yml
```diff
@@ -0,0 +1,50 @@
...
+  - activate test-environment
+
+  # Majority of dependencies can be installed with Anaconda
+  - conda install numpy scipy matplotlib sympy networkx nose h5py pandas
+
+  # Graphviz & PyGraphviz
+  - appveyor DownloadFile "http://graphviz.org/pub/graphviz/stable/windows/graphviz-2.38.zip" -FileName graphviz.zip
+  - 7z x graphviz.zip -y > nul
+  - set "PATH=%cd%\release\lib\release\dll;%PATH%"
+  - pip install https://dl.dropboxusercontent.com/u/6251352/pygraphviz-1.3.1-cp%PYTHON_SVER%-none-win32.whl?dl=1
```
jmuhlich: I found a conda channel offering pygraphviz 1.3.1 for win32/win64: `marufr`. So `conda install -c marufr pygraphviz` ought to do it, then we won't need to maintain this in dropbox.

appveyor.yml
```diff
@@ -0,0 +1,50 @@
+build: false
+
+environment:
+  matrix:
+    - PYTHON_VERSION: 2.7
+      PYTHON_SVER: 27
+      MINICONDA: C:\Miniconda
+    - PYTHON_VERSION: 3.4
```
jmuhlich: Why not test on 3.5 or 3.6 (or both)? Appveyor does offer both:
https://www.appveyor.com/docs/build-environment/#miniconda
Then we could have numpy 1.12 and avoid dropping those doctests.

alubbock: The limiting factor is the pygraphviz binary - I haven't been able to find one for 3.5 or 3.6 on win32, so I had to go with Chris Gohlke's 3.4 build.

appveyor.yml
alubbock: `marufr`'s build is Python 2.7 only, unfortunately.

alubbock:
PySB recently broke compatibility with Kappa 3.5, so that'll need updating to 4 in Appveyor. Need to wait for the Kappa team to release a 4 binary for Windows. There's a pending issue for this: Kappa-Dev/KaSim#394.

appveyor.yml
jmuhlich: ARRGH!  :)

jmuhlich:
OK I guess we will have to let this sit until Kappa releases the v4 windows binary.

PR #265, opened 2017-05-01
Kappa4 interface fixes

bgyori:
This PR introduces updates to the Kappa (KaSim and KaSa) interface in PySB to make simulations and static analysis (generating contact maps and influence maps) compatible with the 4th generation Kappa tools. 
- The executable for static analysis changed from KaSim to KaSa
- The output folders for contact map and influence map have to be specified separately
- The parameterization of KaSim changed, most importantly with respect to specifying the number of time points to simulate vs intervals between returned samples (the -p option of KaSim)
- The output format of KaSim changed and this required changes in reading in the results
- The prerequisites of compiling Kappa changed and these are updated in the Travis config

Note that KaSim doesn't produce an output file if no Observables are declared, resulting in run_simulation to error. This change in KaSim behavior is not addressed here and a solution could be implemented in a follow-up.

### DIFF
--- a/pysb/kappa.py
+++ b/pysb/kappa.py
@@ -408,12 +408,10 @@ def _parse_kasim_outfile(out_filename):
         array with the names of the observables.
     """
     try:
-        with open(out_filename, 'r') as fh:
-            while True:
-                header_line = fh.readline()
-                # Break when the first non-comment line is found
-                if not header_line.startswith('#'):
-                    break
+        fh = open(out_filename, 'r')
+        for header_line in fh:
+            if header_line[0] != '#':
+                break
 
         raw_names = [term.strip() for term in re.split(',', header_line)]
         column_names = []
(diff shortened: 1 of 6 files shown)


pysb/kappa.py
```diff
@@ -138,8 +141,12 @@ def run_simulation(model, time=10000, points=200, cleanup=True,
     fm_filename = base_filename + '_fm.dot'
     out_filename = base_filename + '.out'
 
-    args = ['-i', kappa_filename, '-t', str(time), '-p', str(points),
-            '-o', out_filename]
+    if points == 0:
+        raise ValueError('The number of data points cannot be zero.')
+    plot_period = (1.0 * time / points) if time > 0 else 1.0
```
jmuhlich: I assume the `1.0 *` is to force float division? Prefer `float(time) / points` **or** `from __future__ import division` at the top of the module (I just checked and there's no integer division in this module, in fact there's no other division at all).

pysb/kappa.py
```diff
@@ -402,26 +410,31 @@ def _parse_kasim_outfile(out_filename):
 
     try:
         out_file = open(out_filename, 'r')
+        preamble_len = 2 # First 2 lines are comments
```
jmuhlich: I'd suggest looking for the leading `#` on the comments would be more robust than skipping a fixed number of lines, just in case they change the number of headers (I figure the comment character is unlikely to change). E.g.:

```python
for header_count, header_line in enumerate(out_file):
    if header_line[0] != '#':
        break
# No need to call .strip() on header_line before the re.split line
...
arr = np.loadtxt(... skiprows=header_count ...)
```

Issue #139, opened 2015-09-04
fix global reaction_cache variable in bng.py


jmuhlich:
Pass explicitly to function that uses it.

Issue #267, opened 2017-05-03
Simulator.run modifies the state of the simulator


jmuhlich:
After showing users the new `Simulator` class, I have found they are surprised and dismayed to discover that passing tspan, initials or param_values to `run` causes those values to persist in that simulator instance. I'd like to look at whether we can refactor `run` so it has no side effects. That might be a good time to revisit the overall logic flow in the `initials` and `param_values` getters and setters, as they have grown rather hairy to support a large matrix of use cases.

alubbock:
This is documented in `ScipyOdeSimulator.run`, but I agree it's undesirable and should be changed.

PR #272, opened 2017-05-10
Add BioNetGen Simulator support

alubbock:
This PR adds support for BioNetGen's ODE (Sundials CVODE), Partitioned-leaping algorithm, SSA and NFsim simulators. NFsim in particular requires significant changes, since there's no species list generated. I've also added logging support to the BioNetGen interface and removed the `global` usage in the netfile parser.

I intend to address pysb/pysb#267 (setting initials and param_values in run calls should not be persistent) soon, since network-free is the last major class of simulators for which I'm adding support for now, clearing the way for refactoring the initials support in a later PR.

Resolves: pysb/pysb#139


pysb/simulator/base.py
```diff
@@ -501,43 +532,52 @@ class SimulationResult(object):
...
-            # trajectories is a 3D array, create a list of 2D arrays
-            # This is just a view and doesn't copy the data
-            self._y = [tr for tr in trajectories]
+        if trajectories is None and observables_and_expressions is None:
+            raise ValueError('Need to supply at least one of species '
+                             'trajectories or observables_and_expressions')
+
+        if trajectories is not None and len(trajectories) > 0:
+            # Validate incoming trajectories
+            if hasattr(trajectories, 'ndim') and trajectories.ndim == 3:
```
jmuhlich: Possibly simpler:
```python
if getattr(trajectories, 'ndim', None) == 3:
```

pysb/simulator/base.py
```diff
@@ -501,43 +532,52 @@ class SimulationResult(object):
...
+                # trajectories is a 3D array, create a list of 2D arrays
+                # This is just a view and doesn't copy the data
+                self._y = [tr for tr in trajectories]
+            else:
+                # Not a 3D array, check for a list of 2D arrays
+                try:
+                    if any([tr.ndim != 2 for tr in trajectories]):
```
jmuhlich: No need for `[ ]` around the comprehension, as `any` accepts a generator expression as well.

pysb/simulator/base.py
```diff
@@ -563,38 +603,54 @@ def __init__(self, simulator, tout, trajectories, squeeze=True,
...
-                    self._y[n][:, obs.species] * obs.coefficients).sum(axis=1)
-
-            # expressions
-            sym_dict = dict((k, self._yobs[n][k]) for k in obs_names)
-            sym_dict.update(dict((p.name, param_values[
-                n // self.n_sims_per_parameter_set][i]) for i, p in
-                            enumerate(self._model.parameters)))
-            for i, expr in enumerate(exprs):
-                self._yexpr_view[n][:, i] = expanded_exprs[i](**sym_dict)
+        if observables_and_expressions is not None and len(
```
jmuhlich: Just `if observables_and_expressions:` should be OK here. `None` and empty list are both falsey.

pysb/simulator/bng.py
```diff
@@ -0,0 +1,227 @@
...
+
+        if method not in self._SIMULATOR_TYPES:
+            raise ValueError("Method must be one of " +
+                             str(self._SIMULATOR_TYPES))
+
+        if method == 'hpp':
+            if population_maps is None:
+                raise ValueError('population_maps need to be supplied')
+            if not isinstance(population_maps, collections.Iterable) or not \
+                    all([isinstance(pm, PopulationMap) for pm in
```
jmuhlich: Square brackets not needed. Also in general I feel `any(not isinstance...)` is preferable to `not all(isinstance)` when using generator expressions as the former will short-circuit and stop iterating on the first failure. Obviously with the list comprehension there's no difference because it will produce the entire list first.

pysb/simulator/bng.py
```diff
@@ -0,0 +1,227 @@
...
+                         population_maps]):
+                raise ValueError('population_maps should be a list of '
+                                 'PopulationMap objects')
+
+        # Calculate time intervals and validate
+        if not np.allclose(self.tspan, np.linspace(0, self.tspan[-1],
```
jmuhlich: Something seems off here. Do we really need this check on `tspan`, and if so, why does the exception reference StochKit?

pysb/simulator/bng.py
```diff
@@ -0,0 +1,227 @@
...
+        return SimulationResult(self, tout=tout, trajectories=species_out,
+                                observables_and_expressions=obs_exp_out,
+                                simulations_per_param_set=n_runs)
+
+
+class PopulationMap(object):
+    def __init__(self, complex_pattern, lumping_rate, counter_species=None):
+        try:
+            self.complex_pattern = as_complex_pattern(complex_pattern)
+        except:
```
jmuhlich: Should explicitly catch only `InvalidComplexPatternException`.

pysb/tests/test_simulator_bng.py
```diff
@@ -0,0 +1,115 @@
...
+
+    sim = BngSimulator(model, tspan=np.linspace(0, 1))
+    x = sim.run(n_runs=1, method='nf', seed=_BNG_SEED)
+    observables = np.array(x.observables)
+    assert len(observables) == 50
+
+    A = model.monomers['A']
+    x = sim.run(n_runs=2, method='nf', tspan=np.linspace(0, 1),
+                initials={A(): 100}, seed=_BNG_SEED)
+    print(x.dataframe.loc[0, 0.0])
```
jmuhlich: Testcases shouldn't print anything.

pysb/tests/test_simulator_bng.py
```diff
@@ -0,0 +1,115 @@
...
+    model.reset_equations()
+
+    sim = BngSimulator(model, tspan=np.linspace(0, 1))
+    x = sim.run(n_runs=1, method='ode')
+    assert len(x.expressions) == 50
+    assert len(x.observables) == 50
+
+
+def test_nfsim():
+    # Make sure no network generation has taken place
```
jmuhlich: Does this actually verify that network generation hasn't occurred? I can't see it.

pysb/tests/test_simulator_bng.py
```diff
@@ -0,0 +1,115 @@
...
+
+    A = model.monomers['A']
+    x = sim.run(n_runs=2, method='nf', tspan=np.linspace(0, 1),
+                initials={A(): 100}, seed=_BNG_SEED)
+    print(x.dataframe.loc[0, 0.0])
+    assert np.allclose(x.dataframe.loc[0, 0.0], [100.0, 0.0, 0.0])
+
+
+def test_hpp():
+    # Make sure no network generation has taken place
```
jmuhlich: Same as above.

pysb/bng.py
```diff
@@ -49,14 +52,25 @@ class BngBaseInterface(object):
...
     @abc.abstractmethod
     def __init__(self, model=None, verbose=False, cleanup=False,
-                 output_prefix=None, output_dir=None):
+                 output_prefix=None, output_dir=None,
+                 model_additional_species=None,
+                 model_population_maps=None):
+        self._logger = get_logger(__name__,
+                                  model=model,
+                                  log_level=verbose)
+        if model:
```
alubbock: The Interface can be called with or without a `model` object. Most use cases involve a PySB model which is loaded into BNG, but sometimes it is useful to load a BNGL file directly and use the interface as a "remote control" for BNG without a PySB model. The original motivation was the BNGL importer, which does the latter: https://github.com/pysb/pysb/blob/v1.4.0/pysb/importers/bngl.py#L30-L38

jmuhlich: I mean I get the model present/absent part, but why is it switching on accessing `self._logger.logger` vs. `self._logger`? What even *is* `self._logger.logger`?

pysb/bng.py
```diff
@@ -395,14 +459,15 @@ def execute(self, reload_netfile=False):
                                  cwd=self.base_directory,
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE)
-            if self.verbose:
+            if self.verbose <= logging.DEBUG:
```
jmuhlich: I guess if there are a *lot* of lines there might be a small performance impact for all those calls to `logger.debug`. If that's the thinking, probably just need a comment explaining it.

pysb/bng.py
alubbock: logging.py's `get_logger` creates a `PySBModelLoggerAdapter` if a model is supplied, or a `Logger` otherwise. For most purposes the API for an Adapter and a Logger is similar, but it doesn't have a `getEffectiveLevel()` method. `self._logger.logger` is the underlying Logger object when `self._logger` is actually an adapter, not a logger. I'll add a comment to clarify this.

pysb/bng.py
alubbock: Performance was my original rationale for this, but there'll only be a lot of lines when BNG's verbose options are set, so this option seems redundant. Will remove.

pysb/tests/test_simulator_bng.py
alubbock: This comment relates to the `reset_equations()` line - it doesn't check per se, but it ensures equations don't persist from previous network generation if `robertson` was previously imported. Will move comment and clarify.

pysb/simulator/bng.py
alubbock: It's only NFsim that can only take `t_end` and `n_steps`, not a list of `sample_times`. Will switch to `sample_times` where possible, and limit the check and Exception to NFsim-based simulations. StochKit reference was a mistake from copy/paste; will amend.

pysb/bng.py
```diff
@@ -421,10 +486,65 @@ def action(self, action, **kwargs):
         action_args = self._format_action_args(**kwargs)
 
         # Add the command to the queue
-        self.command_queue.write('\t%s({%s})\n' % (action, action_args))
+        if action_args == '':
```
alubbock: Now implemented in the console interface as well.

alubbock:
@jmuhlich Thanks for the helpful feedback. All implemented except where commented on PR page. On @lh64's advice, I removed the `method="hpp"` option - the user can do HPP simulations with `method="nf"` and supplying a `population_maps` argument.

pysb/bng.py
jmuhlich: Ah, I see. However, `self.verbose` isn't even used after the latest changes to this branch. Can this code just be deleted altogether? :)

FYI, it looks like `LoggerAdapter` does expose `isEnabledFor` (it just propagates to `self.model` which will then work even if `model` itself is another `LoggerAdapter`). In the future we should probably try to implement this sort of thing via `isEnabledFor`.

alubbock: Noted, thanks. Will remove that block.

jmuhlich:
Fixes #287

PR #275, opened 2017-05-15
Add Airspeed Velocity benchmarking support

alubbock:
This PR adds benchmarking support for PySB using [Airspeed Velocity (asv)](https://github.com/spacetelescope/asv), the same benchmarking library that numpy and friends uses. We plan to add more benchmarks over time.


benchmarks/simulator.py
```diff
@@ -0,0 +1,37 @@
...
+import numpy as np
+import timeit
+
+
+class Earm10ODESuite(object):
+    def setup(self):
+        self.nsims = 100
+        self.timer = timeit.default_timer
+        self.model = earm_1_0.model
+        self.parameter_set = np.ones((self.nsims, len(self.model.parameters)))
```
jmuhlich: The initialization to all ones followed by updates with `*=` seems really roundabout, unless I'm missing something obvious. The following seems clearer if the goal is just to expand the parameter values list n times:
```python
self.parameter_set = np.repeat([[p.value for p in self.model.parameters]], self.nsims, axis=0)
```

alubbock: One of those "what the hell was I thinking" moments. Your solution is much better, will use it. Thanks.

PR #276, opened 2017-05-15
Much faster reaction network loading from BNG

alubbock:
Currently, model species, reactions and ODEs are created when loading a `.net` file from BioNetGen. However, many simulators do not need the ODE list, instead using the reactions. As models get large, parsing the `.net` file was taking increasing amounts of time.

ODEs are now generated on demand when the `model.odes` attribute is accessed, saving vast amounts of generation time for large models. Additionally, sympy expressions are only evaluated (simplified) in the final assembly step for each ODE, rather than when every term is added, which vastly speeds things up.

E.g. on the EGFR model limited to 8 iterations of network generation, `generate_equations` previously took 2.7 minutes on my laptop, now takes 26 seconds. For the entire EGFR network model, which is huge at 18,951 species and 415,858 reactions, reading the network file did not complete after 24 hours on my laptop. Now it takes 3.7 minutes to load the species and reactions from the `.net` file, and 3.3 minutes to generate the ODEs, if required.


alubbock:
This PR is based off of #272 and #275, so those two could be merged first for an easier diff.

alubbock:
Closing this PR, which is superseded by #293.

PR #277, opened 2017-05-15
Fix StochKitSimulator for models w/ expressions

alubbock:
* Enable simulation of models with expressions on the `StochKitSimulator` class
* Add an `expand_observables` argument to `Expression.expand_expr()`, which expands Observables into species, reducing code duplication
* The `StochKitExporter` is now much faster for large models
* Reactions which are mass action according to the StochKit definition are now exported as mass action in the StochKit XML. This means that custom propensity functions don't have to be compiled for these reactions, reducing StochKit compile time for large models. It also means that mass action only models can be run using the Windows StochKit Lite, which doesn't have custom propensity function (non mass action) capability.

### DIFF
--- a/pysb/export/stochkit.py
+++ b/pysb/export/stochkit.py
@@ -190,11 +190,12 @@ class StochKitExporter(Exporter):
         # Parameters
         params = etree.Element('ParametersList')
         for p_id, param in enumerate(self.model.parameters):
-            if param.name == 'vol':
-                param.rename('__vol')
+            p_name = param.name
+            if p_name == 'vol':
+                p_name = '__vol'
             p_value = param.value if param_values is None else \
                 param_values[p_id]
-            params.append(self._parameter_to_element(param.name, p_value))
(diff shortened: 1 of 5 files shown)


alubbock:
Also refactored the StochKit export code to make it faster for large models.

pysb/export/stochkit.py
```diff
@@ -181,62 +190,106 @@ def export(self, initials=None, param_values=None):
...
+            reactants = defaultdict(int)
+            products = defaultdict(int)
             # reactants
             for r in rxn["reactants"]:
-                r = "__s%d" % r
-                if r in reactants:
-                    reactants[r] += 1
-                else:
-                    reactants[r] = 1
+                reactants["__s%d" % r] += 1
```
jmuhlich: Good call on the `defaultdict`s.

jmuhlich:
Sorry for the repeated merges from master. I'm trying to review and merge too many PRs at once.

PR #279, opened 2017-05-23
Construct ODE expressions more efficiently

bgyori:
This is a commit by @jmuhlich which very significantly optimizes the construction of equations in a PySB model *after* BNG has run its reaction network generation. Unless there's a reason not to, it would be great to merge this.


alubbock:
I implemented a similar idea in #276.

jmuhlich:
I resubmitted this as #293 with improvements. Closing.

PR #283, opened 2017-06-03
Update BNG version regexp

### LINKED ISSUE #282: Parse netfile errors with BioNetGen 2.3
This regexp doesn't match with BNG 2.3: https://github.com/pysb/pysb/blob/master/pysb/bng.py#L552
since the first line is now
```
# Created by BioNetGen 2.3.0-2.3.0
```

bgyori:
Fixes #282, only changes the regexp to parse the first line of a BNGL net file.


jmuhlich:
The version "2.3.0-2.3.0" was actually a BNG bug, so we shouldn't need this fix.

Issue #13, opened 2011-10-24
ComplexPattern.is_equivalent_to doesn't canonicalize properly


jmuhlich:
The literal site_conditions comparison requires bond numbering to be identical, so some sort of canonicalization of that numbering is necessary.

johnbachman:
Wait, are you sure this is the issue you meant to close with that PR?

jmuhlich:
Oh no. Alex wrote "closes #13" to reference LoLab-VU/pysb #13, not pysb/pysb #13! I don't know how to avoid that from happening...

alubbock:
I'll use the long format, LoLab-VU/pysb#13, in the future. Sorry for the mix up!

PR #285, opened 2017-06-08
Pattern matcher and model test suite

alubbock:
Add a pattern matching engine (pattern.py) and a model test suite built off of that engine (testing/modeltests.py).

# Pattern Matching Engine

The pattern engine uses a graph-based implementation with networkx for checking pattern isomorphism. It can do one off matching (`SimplePatternMatcher` class) or repeated lookups (`SpeciesPatternMatcher`, `ReactionPatternMatcher` and `RulePatternMatcher`) which create lookup caches for fast repeated pattern matching. See the docstring examples in those classes for details. The framework is designed for large numbers of tests against a fixed list of species/reactions/rules (typically, those from a single model). Thus, the functions typically try to "fail fast" (i.e. reject putative matches quickly, where possible, by performing computationally simpler checks first).

The core of the system is a new [networkx](https://networkx.github.io/)-based graph representation of ComplexPatterns (see `ComplexPattern.as_graph()` documentation for implementation details). Patterns can then be checked for exact equality (species equivalence) using a graph isomorphism check, or for subgraph equality (for matching a pattern against species) using a subgraph isomorphism check. This is implemented in `SimplePatternMatcher.match_complex_pattern()`.

ReactionPatterns are matched by checking pairwise combinations of the two ReactionPatterns' sets of ComplexPatterns. If there's a one-to-one match, the match is trivially True, if there are  multiple possible matches, we need to check if a match is possible without reusing the same ComplexPattern. This can be done by encoding the two sets of ComplexPatterns in a bipartite graph and checking for a maximum matching. More details are in the `SimplePatternMatcher.match_reaction_pattern()` docstring.

`SpeciesPatternMatcher` is used for cache-backed comparison of patterns against a list of species (typically, the list of species in a model). Internally, a `_species_cache` is created which creates a Monomer->list of species mapping for rapid lookup. We can then extract the list of monomers in the test pattern, retrieve matching species for each of those monomers in the `_species_cache`, and take the intersection. This allows us to rapidly create a shortlist of candidate matches before doing full graph-based isomorphism comparison (which is computationally expensive).

The `ReactionPatternMatcher` and `RulePatternMatcher` work in a similar vein. These matchers have cache tables mapping Monomer->list of reactions/rules for the reaction and pattern sides separately, as we often want to check just one side of a reaction/rule.

# Model Test Suite

The model test suite allows checking of model assumptions as assertions, e.g. that certain species are generated or not generated, or appear only as reactants or only as products. These may be useful for large models or those with complicated rule sets. See the docstring examples in `TestSuite` for details.

This builds on the pattern matching suite defined above, and can be used interactively or as part of a model's unit tests. The user creates a `TestSuite` instance, and then adds any desired tests on the model's species list, rule set, or reaction network. In principle, these tests could check any other aspect of a model, by implementing new checks as subclasses of `ModelAssertion`. Several checks are implemented "out of the box", e.g. `SpeciesOnlyProduct` (a species appears only on the product side of all reactions), and `AllObservablesInRules` (checks all model observables are specified in at least one rule).

Resolves: pysb/pysb#13

### DIFF
--- a/pysb/testing/modeltests.py
+++ b/pysb/testing/modeltests.py
@@ -1,0 +1,389 @@
+from __future__ import print_function
+import pysb
+import abc
+from pysb.core import SelfExporter
+from pysb.pattern import SpeciesPatternMatcher, ReactionPatternMatcher, \
+    RulePatternMatcher
+
+
+class ModelAssertionFailure(Exception):
+    def __init__(self, assertion, model, message=None):
+        self.assertion = assertion
+        self.model = model
+        self.message = message
+
+    def __repr__(self):
+        base_msg = '%s on model %s failed' % (self.assertion,
+                                              self.model.name)
+        if self.message:
+            return base_msg + ': ' + str(self.message)
+        else:
+            return base_msg
+
+    def __str__(self):
+        return repr(self)
+
+
+class TestSuite(object):
+    """
+    A suite of tests for checking properties of a model
+
+    There are two modes of operation: building a test suite using add() and
+    executing all the tests at once with check_all(), or executing tests
+    immediately with check().
+
+    Examples
+    --------
+
+    Create a test suite for the EARM 1.0 model:
+
+    >>> from pysb.testing.modeltests import TestSuite, SpeciesExists, \
+        SpeciesDoesNotExist
+    >>> from pysb.bng import generate_equations
+    >>> from pysb.examples.earm_1_0 import model
+    >>> ts = TestSuite(model)
+
+    Create variables for model components (not needed for models defined
+    interactively):
+
+    >>> AMito, mCytoC, mSmac, cSmac, L, CPARP = [model.monomers[m] for m in \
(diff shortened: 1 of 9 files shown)


pysb/testing/modeltests.py
```diff
@@ -0,0 +1,389 @@
...
+        if self.message:
+            return base_msg + ': ' + str(self.message)
+        else:
+            return base_msg
+
+    def __str__(self):
+        return repr(self)
+
+
+class TestSuite(object):
```
jmuhlich: How would one use this in an actual test suite? `assert ts.check(...)` ?

alubbock: Yes. In principle, the same assertion could be checked against two test suite instances, e.g. add a Rule, regenerate the network, create a new test suite, and compare `ts1.check(my_assertion)` to `tst2.check(my_assertion)`.

One alternative might be `ts.assert(...)` instead of `assert ts.check(...)`. Do you think that'd be a syntactic improvement?

PR #291, opened 2017-07-21
Fix SimulationResult.all with unicode obs/exprs

### LINKED ISSUE #290: SimulationResult.all errors when observable or expression names are unicode strings
@subkar found another place (`SimulationResult.all`) where we use model component names directly as numpy recarray column names, causing problems when these names are `u''` unicode strings. I think we should be pulling the already "cleaned" names from `self._yobs` and `self._yexpr` rather than recreating them.

alubbock:
This patch fixes `SimulationResult.all` when any observable or expression has a unicode name, and amends the unit tests to check `.all` works.

Fixes #290

### DIFF
--- a/pysb/simulator/base.py
+++ b/pysb/simulator/base.py
@@ -679,10 +679,10 @@ class SimulationResult(object):
             sp_names = ['__s%d' % i for i in range(len(self._model.species))]
             yfull_dtype = zip(sp_names, itertools.repeat(float))
             if len(self._model.observables):
-                yfull_dtype += zip(self._model.observables.keys(),
+                yfull_dtype += zip(self._yobs[0].dtype.names,
                                    itertools.repeat(float))
             if len(self._model.expressions_dynamic()):
-                yfull_dtype += zip(self._model.expressions_dynamic().keys(),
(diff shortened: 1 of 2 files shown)


pysb/simulator/base.py
```diff
@@ -679,10 +679,10 @@ def all(self):
             sp_names = ['__s%d' % i for i in range(len(self._model.species))]
             yfull_dtype = zip(sp_names, itertools.repeat(float))
             if len(self._model.observables):
-                yfull_dtype += zip(self._model.observables.keys(),
+                yfull_dtype += zip(self._yobs[0].dtype.names,
```
jmuhlich: I think this can be simplified to:
```python
yfull_dtype += self._yobs[0].dtype.descr
```

pysb/simulator/base.py
```diff
@@ -679,10 +679,10 @@ def all(self):
...
                                    itertools.repeat(float))
             if len(self._model.expressions_dynamic()):
-                yfull_dtype += zip(self._model.expressions_dynamic().keys(),
+                yfull_dtype += zip(self._yexpr[0].dtype.names,
```
jmuhlich: Simplify same as above.

PR #292, opened 2017-07-24
Save and load SimulationResults to HDF5 files

alubbock:
This PR implements `SimulationResult.save()` and `.load()` functions, using an HDF5-based file format. HDF5 is a hierarchical, binary storage format well suited to storing matrix-like data. This implementation uses the h5py package.

The file format stores not only the simulation trajectories, but also metadata relevant to archiving, such as the model itself, simulation settings, parameter and initial condition values, simulation date/time, and PySB version.

The goals are to aid reproducibility and to facilitate exchange of simulation data.

### DIFF
--- a/pysb/simulator/base.py
+++ b/pysb/simulator/base.py
@@ -7,12 +7,19 @@ import numbers
 from pysb.core import MonomerPattern, ComplexPattern, as_complex_pattern, \
                       Component
 from pysb.logging import get_logger, EXTENDED_DEBUG
-import logging
+import pickle
+from pysb import __version__ as PYSB_VERSION
+from datetime import datetime
+import dateutil.parser
 
 try:
     import pandas as pd
 except ImportError:
     pd = None
+try:
+    import h5py
+except ImportError:
+    h5py = None
 
 
 class SimulatorException(Exception):
@@ -97,6 +104,7 @@ class Simulator(object):
         self.initials = initials
         self._params = None
         self.param_values = param_values
+        self._init_kwargs = kwargs
 
     @property
     def model(self):
@@ -462,6 +470,13 @@ class SimulationResult(object):
         deterministic simulators (e.g. ODE), but with stochastic simulators
         multiple trajectories per parameter/initial condition set are often
         desired.
+    model: pysb.Model
+    initials: numpy.ndarray
+    param_values: numpy.ndarray
+        model, initials, param_values are an alternative constructor
+        mechanism used when loading SimulationResults from files (see
+        :func:`SimulationResult.load`). Setting just the simulator argument
+        instead of these arguments is recommended.
 
     Examples
     --------
@@ -533,15 +548,33 @@ class SimulationResult(object):
     13.333333  0.000002    4.999927
     """
     def __init__(self, simulator, tout, trajectories=None,
-                 observables_and_expressions=None,
-                 squeeze=True,
-                 simulations_per_param_set=1):
-        simulator._logger.debug('SimulationResult constructor started')
+                 observables_and_expressions=None, squeeze=True,
+                 simulations_per_param_set=1, model=None, initials=None,
+                 param_values=None):
+        if simulator:
+            simulator._logger.debug('SimulationResult constructor started')
+            self._param_values = simulator.param_values
+            try:
+                self._initials = simulator.initials
+            except SimulatorException:
+                # Network free simulations don't have initials list, only dict
+                self._initials = simulator.initials_dict
+            self._model = simulator._model
+            self.simulator_class = simulator.__class__
+            self.simulator_kwargs = simulator._init_kwargs
+        else:
+            self._param_values = param_values
+            self._initials = initials
+            self._model = model
+            self.simulator_class = None
+            self.simulator_kwargs = None
+
         self.squeeze = squeeze
         self.tout = np.asarray(tout)
         self._yfull = None
-        self._model = simulator._model
         self.n_sims_per_parameter_set = simulations_per_param_set
+        self.pysb_version = PYSB_VERSION
+        self.creation_date = datetime.now()
 
         if trajectories is None and observables_and_expressions is None:
             raise ValueError('Need to supply at least one of species '
@@ -594,7 +627,7 @@ class SimulationResult(object):
                     try:
                         name_list[i] = name.encode('ascii')
                     except UnicodeEncodeError:
-                        error_msg = 'Non-ASCII compatible' + \
+                        error_msg = 'Non-ASCII compatible ' + \
                                     '%s names not allowed' % name_type
                         raise ValueError(error_msg)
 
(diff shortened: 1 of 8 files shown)


pysb/tests/test_simulationresult.py
```diff
@@ -37,3 +40,113 @@ def test_simres_dataframe():
...
+    # NFsim without expressions
+    nfsim1 = BngSimulator(robertson.model)
+    nfres1 = nfsim1.run(n_runs=2, method='nf', tspan=np.linspace(0, 1))
+
+    # NFsim with expressions
+    nfsim2 = BngSimulator(expression_observables.model)
+    nfres2 = nfsim2.run(n_runs=1, method='nf', tspan=np.linspace(0, 100, 11))
+
+    with tempfile.NamedTemporaryFile() as tf:
+        if os.name == 'nt':
```
jmuhlich: Is there any reason not to close it unconditionally, regardless of OS? Do Mac/Linux have a problem if it's closed?

pysb/simulator/base.py
```diff
@@ -735,15 +782,230 @@ def observables(self):
...
+            If they are not included, they can be recreated from the model
+            and species trajectories when loaded back into PySB, but you may
+            wish to include them for use with external software, or if you
+            have complex expressions which take a long time to compute.
+        """
+        if h5py is None:
+            raise Exception('Please "pip install h5py" for this feature')
+
+        if self._y is None and not include_obs_exprs:
+            raise ValueError('This SimulationResult has no trajectories - '
```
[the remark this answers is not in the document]
alubbock: A warning works. Seems unlikely anyone would want to save an empty set of results, but it shouldn't be outright forbidden by an Exception.

pysb/tests/test_simulationresult.py
alubbock: Closing the file straight away on Mac/Linux as well as Windows does not cause a problem, but the file gets closed at the end of the `with` block anyway. I'll remove the `os.name` test anyway for cross-platform consistency.

pysb/simulator/base.py
```diff
@@ -533,15 +559,35 @@ class SimulationResult(object):
...
+            self.init_kwargs = {}
+            self.run_kwargs = {}
+
         self.squeeze = squeeze
         self.tout = np.asarray(tout)
         self._yfull = None
-        self._model = simulator._model
         self.n_sims_per_parameter_set = simulations_per_param_set
+        self.pysb_version = PYSB_VERSION
+        self.creation_date = datetime.now()
```
jmuhlich: Minor nitpick, but this isn't just a date. Would `timestamp` be an acceptable name?

pysb/simulator/base.py
```diff
@@ -735,15 +783,230 @@ def observables(self):
...
+
+        if self._y is None and not include_obs_exprs:
+            warn('This SimulationResult has no trajectories - '
+                 'you will need to set include_obs_exprs=True if '
+                 'you wish to save observables and expressions')
+
+        if group_name is None:
+            group_name = self._model.name
+        if dataset_name is None:
+            dataset_name = self.simulator_class.__name__
```
jmuhlich: I find this an odd default for the dataset name. How about something neutral like `'result'`?

pysb/simulator/base.py
```diff
@@ -735,15 +783,230 @@ def observables(self):
...
+                    raise ValueError('SimulationResult model has name "{}", '
+                                     'but the model in HDF5 file group "{}" '
+                                     'has name "{}"'.format(self._model.name,
+                                                            group_name,
+                                                            model.name))
+
+            # Create the result dataset, which is actually a nested HDF group
+            dset = grp.create_group(dataset_name)
+            if self._y is not None:
+                dset.create_dataset('trajectories', data=self._y,
```
jmuhlich: I suggest adding `shuffle=True` to all of the `create_dataset` calls, but most of all this one.

pysb/simulator/bng.py
```diff
@@ -74,7 +74,8 @@ def run(self, tspan=None, initials=None, param_values=None, n_runs=1,
         """
         super(BngSimulator, self).run(tspan=tspan,
                                       initials=initials,
-                                      param_values=param_values
+                                      param_values=param_values,
+                                      _run_kwargs=locals()
```
jmuhlich: This seems hacky. At the very least, `Simulator.run` needs to document the contract that subclasses need to follow when overriding it.

An alternative would be to make all `run` methods take and propagate `**kwargs`, and write a helper to manage setting default values, but of course the `run` signatures will be less useful without the explicit arg names and defaults.

pysb/simulator/base.py
jmuhlich: As a point of reference I ran 5000 simulations of earm_1_0 with slightly perturbed initial conditions, resulting in a `trajectories` dataset of 1000 x 361 x 58, which is 838MB raw. Adding shuffle in addition to gzip saved 9% on the final file size (726MB -> 665MB).

pysb/simulator/base.py
```diff
@@ -735,15 +783,230 @@ def observables(self):
...
+                model = pickle.loads(grp['_model'][()])
+                if model.name != self._model.name:
+                    raise ValueError('SimulationResult model has name "{}", '
+                                     'but the model in HDF5 file group "{}" '
+                                     'has name "{}"'.format(self._model.name,
+                                                            group_name,
+                                                            model.name))
+
+            # Create the result dataset, which is actually a nested HDF group
+            dset = grp.create_group(dataset_name)
```
jmuhlich: I had the idea that it might be helpful to allow user-specified attributes on this group, to store notes about each simulation run. Nothing fancy, just numbers and strings I think. Maybe it's an explicit argument to `save` or maybe it's set as a property on `self` and pulled from there.

pysb/simulator/bng.py
alubbock: I've updated the `Simulator.run()` documentation to reflect the suggested contract with the `_run_kwargs` argument. If it's not supplied, a warning is now generated.

pysb/simulator/base.py
alubbock: Good idea. I implemented this as a `custom_attrs` dict on `SimulationResult`, where the values get pickled upon saving to HDF5.

jmuhlich: Well, actual HDF5 attributes would be more helpful, so the values are accessible from tools other than pysb (or at least other than Python). That's why I suggested restricting the values, so they can be stored directly as attributes. The keys in `custom_attrs` would of course also be restricted to strings.

I realize now that some sort of namespacing would be needed to be able to store each element in its own attribute. Could either keep them at this level by prefixing the keys, or add a new empty child group just to hang the attributes on.

alubbock: The `custom_attrs` dictionary entries are already stored separately as HDF5 attributes, as you suggest. Each key in `custom_attrs` is prefixed with "usrattr_" when saved to HDF5. Each value is pickled to allow storing arbitrary data types, but the strings are still human readable. E.g. the pickled string `NFsim without expressions` is stored as `S'NFsim without expressions'\np0\n.`. Would be fairly easy for non-Python software to extract the string, and the added flexibility of being able to store multiple object types seems worth having, no?

jmuhlich:
@alubbock I pushed my work on storing `custom_attrs` natively where possible to [jmuhlich/pysb:simres_hdf5](https://github.com/LoLab-VU/pysb/compare/simres_hdf5...jmuhlich:simres_hdf5). Note that it's two commits, the first is some general Py3 fixups to the save/load code.

jmuhlich:
Thanks for being patient with my slow review! Nice work on this feature.

PR #293, opened 2017-07-24
Construct ODE expressions more efficiently

jmuhlich:
This is an improvement on #279 which @bgyori submitted on my behalf. The approach is now aligned more closely with @alubbock 's #276 and I improved the performance even more using some tricks I learned from that code.

The key difference here is that the ODE expressions are lazily evaluated rather than pre-cached in full. I found that consumers of `model.odes` tend to just iterate once over the list, so for models with very complex equations this can save some memory (particularly the ones we've been automatically constructing in our indra project).

There is one thing that's pre-cached upon the first access of an element of `odes` -- the stoichiometry matrix for the system (as a sparse matrix of `int`s). This is fairly quick to compute and takes very little memory. Also it can be very useful for certain analyses in and of itself.

Also this implementation benchmarks faster than #276 for our large indra-generated models, although the current ASV benchmark on `egfr_extended` doesn't show any difference.


pysb/core.py
```diff
@@ -1671,6 +1689,32 @@ def rename(self, c, new_name):
...
+        self.model = model
+
+    def __getitem__(self, key):
+        if not self.model.species:
+            raise IndexError('list index out of range')
+        if isinstance(key, slice):
+            return [self[k] for k in range(*key.indices(len(self)))]
+        sr = self.model.stoichiometry_matrix[key]
+        terms = [self.model.reactions[i]['rate'] * v
+                 for i, v in zip(sr.indices, sr.data)]
```
alubbock: Replacing this statement with
```python
terms = [sympy.Mul(self.model.reactions[i]['rate'], v, evaluate=False)
         for i, v in zip(sr.indices, sr.data)]
```
speeds the `time_egfr_equations_max_iter_8` benchmark up by around 10% for me.

jmuhlich: Can't argue with that!

jmuhlich: However, it slowed down time_scipy_lsoda for me. It seems to be due to the LHS code having expressions like `+ (x* y* z)*-1` rather than `-x*y*z` combined with my more recent version of scipy which lacks weave. The python RHS evaluator evaluates the expressions directly, whereas in the weave one the C compiler is really good at simplifying them so there's no added overhead. Will need to think on this one a bit more.

alubbock: I've just submitted PR #294, which replaces Python RHS `eval` with `lambdify` and has experimental (optional, off by default) theano support in ScipyOdeSimulator. It's 30% faster on the time_scipy_lsoda benchmark for me (compared with current master 553eadc) and the above change seems to make it slightly faster, not slower, if that helps.

jmuhlich: OK, let's do this then.

PR #294, opened 2017-07-24
ScipyOdeSimulator lambdify and theano support

alubbock:
This PR replaces the previous `eval` mechanism for ODE evaluation in `ScipyOdeSimulator` with `lambdify`, allowing some extra rate laws which would previously fail, e.g. `Piecewise` expressions (see new unit test `test_piecewise_expression`). It is >30% faster on the time_scipy_lsoda benchmark for me vs current master (553eadc).

Experimental [theano](http://deeplearning.net/software/theano/) support is also added, and can be enabled using a `use_theano=True` argument to the `ScipyOdeSimulator` constructor. Theano has some interesting capabilities like GPU offloading which might help with large, complex expression sets, but it needs more testing and benchmarking to establish which simulation scenarios it's appropriate for.

The PR also slightly cleans up the BNG generator code by subclassing sympy's `StrPrinter`.

### DIFF
--- a/pysb/generator/bng.py
+++ b/pysb/generator/bng.py
@@ -2,6 +2,8 @@ import inspect
 import warnings
 import pysb
 import sympy
+from sympy.printing import StrPrinter
+from sympy.logic.boolalg import BooleanTrue
 
 # Alias basestring under Python 3 for forwards compatibility
 try:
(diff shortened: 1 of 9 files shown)


pysb/generator/bng.py
```diff
@@ -249,14 +251,31 @@ def warn_caller(message):
         module = inspect.getmodule(caller_frame)
     warnings.warn(message, stacklevel=stacklevel)
 
+
+class BngPrinter(StrPrinter):
+    def __init__(self, **settings):
+        super(BngPrinter, self).__init__(settings)
+
+    def _print_Piecewise(self, expr):
+        if type(expr.args[-1][1]) != BooleanTrue:
```
alubbock: I thought so too, however it turns out `BooleanTrue` has type `sympy.core.singleton.Singleton`, so the proposed comparison fails.

alubbock:
I replaced `scipy.weave` instances package-wide with `weave`. I couldn't add `weave` to setup.py as this fails on Python 3, but I added it as a conditional install in .travis.yml.

pysb/generator/bng.py
jmuhlich: Weird, `isinstance` does work for me in some minimal test cases with a `Piecewise` object. Anyway, with a true singleton you can use `is` against the singleton instance:
```python
if expr.args[-1][1] is not sympy.true:
```
Then you can also delete the import of `BooleanTrue` above.

alubbock:
The check against `sympy.true` works; I've pushed it in a new commit. Thanks!

PR #297, opened 2017-09-18
Fix link rot in appveyor config

alubbock:
The download links for pygraphviz were broken due to the discontinuation of public Dropbox folders, and the link for BioNetGen was updated on the remote server to point to a newer version than desired for PySB (2.3.0 vs 2.2.6). These links have been updated to fix our appveyor continuous integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ before_install:
 - chmod +x miniconda.sh
 - bash miniconda.sh -b -p $HOME/miniconda
 - export PATH="$HOME/miniconda/bin:$PATH"
-- conda update --yes conda
+# Install conda 4.3.21 to avoid conda/weave issue (see #297 discussion)
+- conda install --yes conda==4.3.21
 - conda install --yes -c omnia python="$TRAVIS_PYTHON_VERSION"
     numpy scipy matplotlib sympy networkx nose h5py pexpect pandas pygraphviz=1.3 theano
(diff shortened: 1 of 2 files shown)


jmuhlich:
The Travis py27 build is failing due to one failure, I think it's doctests on integrate.py.  The exception is from weave but it's not so helpful... "ValueError: The '' compiler was not found." The triggering code is my setup_module function for integrate.py which calls ScipyOdeSimulator_test_inline, but maybe we can delete that now that #236 suppresses the compiler noise.

jmuhlich:
Nope, we still need the setup_module call, as the message to be suppressed isn't from the compiler, but rather internal to distutils.

alubbock:
@jmuhlich I managed to reproduce this build failure on a Linux box. Looks like this is a bug is due to a [build error with anaconda](https://stackoverflow.com/questions/46450912/unable-to-execute-x86-64-conda-cos6-linux-gnu-gcc-no-such-file-or-directory) - weave is searching, and not finding, a compiler called `x86_64-conda_cos6-linux-gnu-gcc`. 

I've pushed a workaround to use the previous patch release of conda (4.3.21) instead of the current one with the issue described above (4.3.27). The [issue is being tracked by conda](https://github.com/conda/conda/issues/6030) and I've subscribed to updates. When it's fixed we can unpin from the 4.3.21 release.

PR #299, opened 2017-09-30
Improve control over weave.inline console output

jmuhlich:
Our logging framework now controls the various output produced by weave.inline
and the C compiler. Due to some misbehavior within distutils, we need to
dynamically monkey-patch its internal logging framework to achieve this.
C compiler warning messages now become visible at log level DEBUG.

Due to these changes, the nose setup_module function in pysb.integrate is no
longer needed and has been removed.

Also an unrelated bug in ScipyOdeSimulator's Jacobian calculation was fixed.

### DIFF
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -152,7 +152,7 @@ class ScipyOdeSimulator(Simulator):
         # logging system, the threshold must be lower than DEBUG or else the
         # Nose logcapture plugin will cause the warnings to be shown and tests
         # will fail due to unexpected output.
-        if True or not self._logger.isEnabledFor(EXTENDED_DEBUG):
+        if not self._logger.isEnabledFor(EXTENDED_DEBUG):
             extra_compile_args.append('-w')
 
         if self._use_inline and not use_theano:
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -5,7 +5,6 @@ try:
     from weave import inline as weave_inline
     import weave.build_tools
     import distutils.errors
-    import distutils.log
 except ImportError:
     weave_inline = None
 try:
@@ -23,6 +22,7 @@ from pysb.logging import get_logger, EXTENDED_DEBUG
 import logging
 import itertools
 import contextlib
+import importlib
 
 
 class ScipyOdeSimulator(Simulator):
@@ -147,8 +147,12 @@ class ScipyOdeSimulator(Simulator):
         self._test_inline()
 
         extra_compile_args = []
-        # Inhibit weave C compiler warnings unless log level <= DEBUG.
-        if not self._logger.isEnabledFor(logging.DEBUG):
+        # Inhibit weave C compiler warnings unless log level <= EXTENDED_DEBUG.
+        # Note that since the output goes straight to stderr rather than via the
+        # logging system, the threshold must be lower than DEBUG or else the
+        # Nose logcapture plugin will cause the warnings to be shown and tests
+        # will fail due to unexpected output.
+        if True or not self._logger.isEnabledFor(EXTENDED_DEBUG):
             extra_compile_args.append('-w')
 
         if self._use_inline and not use_theano:
@@ -173,7 +177,7 @@ class ScipyOdeSimulator(Simulator):
 
             # Call rhs once just to trigger the weave C compilation step while
             # asserting control over distutils logging.
-            with self._override_distutils_logging:
+            with self._patch_distutils_logging:
                 rhs(0.0, self.initials[0], self.param_values[0])
 
         else:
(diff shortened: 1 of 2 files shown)


pysb/simulator/scipyode.py
```diff
@@ -191,10 +205,9 @@ def rhs(t, y, p):
         # put together the sensitivity matrix)
         jac_fn = None
         if self._use_analytic_jacobian:
-            species_names = ['__s%d' % i for i in
-                             range(len(self._model.species))]
-
-            jac_matrix = ode_mat.jacobian(species_names)
+            species_symbols = [sympy.Symbol('__s%d' % i)
```
jmuhlich: This is the jacobian fix. `sympy.Matrix.jacobian` failed when passed strings for the differentiation variables, so I generated actual `Symbol` objects.

pysb/simulator/scipyode.py
```diff
@@ -235,21 +246,24 @@ def jacobian(t, y, p):
...
+                with self._override_distutils_logging:
+                    jacobian(0.0, self.initials[0], self.param_values[0])
+
             else:
                 jac_eqs_py = sympy.lambdify(self._symbols, jac_matrix, "numpy")
 
                 def jacobian(t, y, p):
                     return jac_eqs_py(*itertools.chain(y, p))
 
-            # Initialize the jacobian argument to None if we're not going to
```
jmuhlich: These seem like orphans from a refactoring.

jmuhlich:
@alubbock Please test this on your Mac with Clang to determine whether that `####` warning message is correctly handled. You should have to set the logging level to `INFO` or lower to see it now.

pysb/simulator/scipyode.py
```diff
@@ -384,3 +410,57 @@ def run(self, tspan=None, initials=None, param_values=None):
...
+        logger.debug('restoring distutils and numpy.distutils logging')
+        # Restore everything we overwrote.
+        for (module, name), value in saved_symbols.items():
+            setattr(module, name, value)
+
+
+class _DistutilsProxyLoggerAdapter(logging.LoggerAdapter):
+    """A logging adapter for the distutils logging patcher."""
+    def process(self, msg, kwargs):
+        return '(from %s) %s' % (self.extra['module'], msg), kwargs
```
jmuhlich: This ends up logging the distutils.log messages under pysb.simulator.scipyode, with a note about which distutils module the call actually originated from. Should we instead create Loggers for those modules directly, which will require some extra logger setup code on our part?

pysb/simulator/scipyode.py
```diff
@@ -143,7 +146,16 @@ def __init__(self, model, tspan=None, initials=None, param_values=None,
...
         self._test_inline()
 
-        if self._use_inline:
+        extra_compile_args = []
+        # Inhibit weave C compiler warnings unless log level <= EXTENDED_DEBUG.
+        # Note that since the output goes straight to stderr rather than via the
+        # logging system, the threshold must be lower than DEBUG or else the
+        # Nose logcapture plugin will cause the warnings to be shown and tests
+        # will fail due to unexpected output.
+        if True or not self._logger.isEnabledFor(EXTENDED_DEBUG):
```
alubbock: This `if` statement is always `True`

pysb/simulator/scipyode.py
```diff
@@ -211,9 +228,7 @@ def jacobian(t, y, p):
                     return jacmat
 
             elif self._use_inline:
-                self.jac = np.zeros(
-                    (len(self._model.odes), len(self._model.species)))
-                # Next, prepare the stringified Jacobian equations
+                # Prepare the stringified Jacobian equations.
```
alubbock: Wasn't `self.jac` technically part of the public API, although undocumented?

pysb/simulator/scipyode.py
alubbock: I think the current approach is sound. It's more useful to know the message came through a Simulator than just seeing that it came from distutils.

pysb/simulator/scipyode.py
jmuhlich: No, this isn't even the symbolic Jacobian matrix. It's just a local scratch space where the `jacobian` function (analogous to the `rhs` function) stores the numeric evaluation of the Jacobian on each invocation, as an optimization over allocating a new array on each call. It has no use to external code. The same holds for `ydot` which I also cleaned up. The original implementation stored these scratch spaces as attributes of `self` and then assigned them to local variables within the integrator callback functions, but I realized they could just be locals in `__init__` and referenced directly in the callbacks. They aren't usable by external code because integrator calls to `rhs` and `jacobian` don't correspond with the currently requested time step in any way.

alubbock: My suggestion was less about the semantics/usability of the variables than the syntax - instance variables without leading underscores are suggestive of a public API. I think they're OK to drop in this case, but we should be careful about this in new code (and preferably, declare all instance variables in `__init__`; at least non-underscored ones).

jmuhlich: Gotcha, agreed.

PR #304, opened 2017-10-12
option for total rate bionetgen keyword

rasi:
The tests for Python 2.7 ran fine, but one for Python 3.5 did not. I don't see any Python 2 specific code in my change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ before_install:
 - chmod +x miniconda.sh
 - bash miniconda.sh -b -p $HOME/miniconda
 - export PATH="$HOME/miniconda/bin:$PATH"
-- conda update --yes conda
+# Install conda 4.3.21 to avoid conda/weave issue (see #297 discussion)
+- conda install --yes conda==4.3.21
 - conda install --yes -c omnia python="$TRAVIS_PYTHON_VERSION"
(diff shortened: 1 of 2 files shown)


alubbock:
Hi @rasi, thanks for your contribution to PySB! Adding the `total_rate` keyword to PySB would be great. Two points that ideally should be rectified before we merge this in:

1. Personally I haven't used the `TotalRate` BNG keyword, but the [BNG release notes](http://bionetgen.org/index.php/Release_Notes#NFsim_compatibility) state it's only used for NFsim. The PySB documentation for the new keyword should reflect this (also, there's a typo in "macroscopic" in the docstring). Does BNG's `generate_network` fail when this keyword is present, or just ignore it? Would be good to include this information in the documentation, too.
2. We like all new code to have unit tests to ensure that it functions correctly, now and in the future. Do you have a minimal example model using `total_rate` that could be used as a test? If possible, please could you add such an example testing the keyword as a new function to pysb/tests/test_bng.py?

Regarding Python 3 compatibility, PySB needs to be specifically built for Python 3 in order for automated converters to apply (e.g. converting the `<>` rule operator, removed in Python 3, to `!=`). Fortunately, the `nose` library has this built in: just run `python3 setup.py nosetests`. You'll see that it passes - your code is already Python 3 compatible.

Thanks again, and please do get in touch if you have any questions about the unit tests, or anything else. Always good to see new contributors!

Issue #302, opened 2017-10-10
is_concrete false positives


jmuhlich:
MonomerPattern.is_concrete will return True if a state site participates in a bond, regardless of whether its state is defined. For example:

```python
Monomer('A', ['s'], {'s': ['x']})
assert not (A(s=1) % A(s=1)).is_concrete()  # This should succeed, but does not.
```
Also it should return False if `ANY` or `WILD` bonds are present.

PR #307, opened 2017-10-19
Fixes and tests for ComplexPattern.is_concrete()

alubbock:
Fixes: #302

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -18,11 +18,9 @@ except NameError:
     from imp import reload
 try:
     basestring
-    STRING_TYPES = (str, unicode)
 except NameError:
     # Under Python 3, do not pretend that bytes are a valid string
     basestring = str
-    STRING_TYPES = (basestring)
     long = int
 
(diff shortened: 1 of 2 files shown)


pysb/core.py
```diff
@@ -404,8 +406,28 @@ def is_site_concrete(self):
         Return a bool indicating whether the pattern is 'site-concrete'.
 
         'Site-concrete' means all sites have specified conditions."""
-        # assume __init__ did a thorough enough job of error checking that this is is all we need to do
-        return len(self.site_conditions) == len(self.monomer.sites)
+        if len(self.site_conditions) != len(self.monomer.sites):
+            return False
+        for site_name, site_val in self.site_conditions.items():
+            if isinstance(site_val, STRING_TYPES):
```
jmuhlich: Could you use `basestring` directly here instead of defining `STRING_TYPES`?

alubbock: Gah, indeed. Fixed.

Issue #271, opened 2017-05-09
all lib2to3 fixers erroneously applied on build, not just fix_ne


jmuhlich:
As of our Python 3 support in #124, we should have only Python-3-compatible source code committed to the repository with one exception: The use of `<>` in example models. setup.py has some logic to apply only the `fix_ne` lib2to3 fixer at build time to deal with this one issue on py3 builds. However it appears that two related issues have arisen: Non-Python-3-compatible code is being committed, and now all of the lib2to3 fixers are being run. I had expected CI to catch Python 3 compatibility issues, but lib2to3 is cleaning them all up before running tests.

We'll need to determine why all the fixers are now running and again disable everything but `fix_ne`, then fix up all py3 problems in our code.

jmuhlich:
#312 will resolve this, but once #309 lands we should drop the 2to3 fixers entirely.

alubbock:
Implemented in #312 (#309 is already merged). We now have native Python 3 compatibility, no fixers required.

PR #312, opened 2017-11-03
Fix some Python 2/3 compatibility issues

bgyori:
This PR fixes some Python 2/3 compatibility problems in an attempt to solve #271. Note that currently
- The `pysb.macros.assemble_chain_sequential_base` doctest is failing in Python 3
- The `pysb.macros.chain_species_base` doctest is failing in Python 3

these will need to be diagnosed and fixed before merging.

I found that the 2to3 issue mentioned in #271 was due to setup.py, where ```
cmdclass.update(versioneer.get_cmdclass())``` overwrote the `build_py` entry of `cmdclass` causing _all_ 2to3 fixes to be applied. This PR first gets the versioneer cmdclasses and then overwrites the `build_py` entry with the custom `build_py` class defined in setup.py. This fixes the issue with 2to3 but might have side effects relevant to versioning that I am not aware of.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -668,7 +668,7 @@ class ComplexPattern(object):
                 bond_num = None
                 if state_or_bond is WILD:
                     continue
-                elif isinstance(state_or_bond, (str, unicode)):
+                elif isinstance(state_or_bond, basestring):
                     state = state_or_bond
                 elif isinstance(state_or_bond, collections.Iterable) and len(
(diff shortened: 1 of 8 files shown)


jmuhlich:
Once we merge #309 to replace `<>` with `|`, we won't need `2to3` at all!

jmuhlich:
Oh, nice sleuthing on the `build_py` problem!

PR #314, opened 2017-11-13
Unpersist Simulator.run() arguments

alubbock:
Previously, tspan, initials, and param_values supplied to
Simulator.run() calls persisted to future run calls, which seems
unintuitive. This commit makes these arguments persist only for the
duration of the run() call. It also refactors some of the more
convoluted logic for setting initials/param_values and adds a few
unit tests.

Fixes: #267

### DIFF
--- a/pysb/simulator/base.py
+++ b/pysb/simulator/base.py
@@ -5,7 +5,7 @@ import sympy
 import collections
 import numbers
 from pysb.core import MonomerPattern, ComplexPattern, as_complex_pattern, \
-                      Component
+                      Parameter, Expression
 from pysb.logging import get_logger, EXTENDED_DEBUG
 import pickle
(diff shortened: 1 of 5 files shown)


pysb/simulator/base.py
```diff
@@ -107,18 +107,52 @@ def __init__(self, model, tspan=None, initials=None,
...
+    def _num_sims_calc(initials_or_params):
+        """ Calculate number of simulations implied by initials or param
+        values """
+        if initials_or_params is None:
+            return None
+
+        if isinstance(initials_or_params, np.ndarray):
+            return len(initials_or_params)
+
+        try:
```
jmuhlich: How about `next(iter(initials_or_params.values()))` as a single 2/3-compatible line?

alubbock: Implemented.

PR #319, opened 2017-11-22
Add warning if using ScipyOdeSimulator w/o weave

alubbock:
I'm finding that many users of PySB are not aware they can install the
additional "weave" package for a significant speed boost to the
ScipyOdeSimulator. This PR adds a warning if pure Python (lambdify)
mode is being used, to instruct users they may wish to install
weave or use the experimental theano mode.


alubbock:
Superseded by #320

PR #320, opened 2017-11-28
Cython support for ScipyOdeSimulator

alubbock:
Adds Cython support to ScipyOdeSimulator. This is useful because our
current ODE compilation backend (weave) does not support Python 3;
Cython does.

The ODE compilation system backend can be specified by a new
`eqn_mode` argument: `cython`, `weave`, `theano` or `python`.
If left as None (default), the system will try to use `weave`,
then `cython`, then `python`.
If the auto-detect falls back on `python` mode, a warning is shown
that the simulation may be slow, and recommends the user install
`cython` or `weave`.


alubbock:
I actually found weave to be faster than Cython in some scenarios, although the difference was small. Perhaps we could keep weave support for one release (with a deprecation warning) to see if we can tweak Cython using head to head comparisons?

Issue #90, opened 2014-06-24
MonomerPattern doesn't verify given site states


jmuhlich:
MonomerPattern.__init__ should verify that any strings (states) given in site_conditions are allowable for that monomer. E.g. if I define this monomer:

```
Monomer('A', ['x'], {'x': ['foo', 'bar']})
```

Then this should raise an exception:

```
A(x='lemon')
```

because 'lemon' was not specified in the definition of 'A' as an allowable state for site 'x'.

PR #322, opened 2017-11-29
Verify site states in MonomerPatterns

alubbock:
`MonomerPattern.__init__` should verify that any strings (states) given
in site_conditions are allowable for that monomer. E.g. for this
Monomer:

    Monomer('A', ['x'], {'x': ['spam', 'eggs']})

This now raises an exception:

    A(x='foo')

Fixes: #90

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -302,6 +302,15 @@ class Monomer(Component):
         return value
 
 
+def _check_state(monomer, site, state):
+    """ Check a monomer site allows the specified state """
+    if state not in monomer.site_states[site]:
+        args = state, monomer.name, site, monomer.site_states[site]
+        template = "Invalid state choice '{}' in Monomer {}, site {}. Valid " \
+                   "state choices: {}"
+        raise ValueError(template.format(*args))
+
+
 class MonomerPattern(object):
 
     """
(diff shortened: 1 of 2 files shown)


pysb/core.py
```diff
@@ -337,6 +337,16 @@ class MonomerPattern(object):
     state for that site, i.e. \"don't write, don't care\".
 
     """
+    @staticmethod
+    def _check_state(monomer, site, state):
+        if state not in monomer.site_states[site]:
+            raise ValueError("Invalid state choice \"{}\" in Monomer {}, "
```
jmuhlich: Can you reformat this? Maybe something like this:
```python
args = state, monomer.name, site, monomer.site_states[site]
template = "Invalid state choice '{}' in Monomer {}, site {}. Valid state choices: {}"
raise ValueError(template.format(args))
```

jmuhlich: Can you reformat this? Maybe something like this:
```python
args = state, monomer.name, site, monomer.site_states[site]
template = "Invalid state choice '{}' in Monomer {}, site {}. Valid state choices: {}"
raise ValueError(template.format(args))
```

pysb/core.py
```diff
@@ -359,11 +369,13 @@ def __init__(self, monomer, site_conditions, compartment):
                  all(isinstance(s, int) for s in state):
                 continue
             elif isinstance(state, basestring):
+                self.__class__._check_state(monomer, site, state)
```
jmuhlich: This doesn't really read well as a staticmethod. Can it just be a top-level underscore-prefixed function?

PR #325, opened 2017-12-04
Update biomodels download URL

alubbock:
The previous EBI link stopped working, which was breaking the unit
tests. I've updated it to the Caltech mirror.


jmuhlich:
Seems to be working now... Do we believe the Caltech mirror is going to be more stable than the EBI one, in the long term?

jmuhlich:
Actually I just realized the solution for the tests is to mock out the biomodels download. We don't need to test that the biomodels website is working, we only need to test our code that sends the request and parses the response.

alubbock:
I added both Biomodels mirrors as a switchable option, and used a mock to use a local SBML file from BioNetGen for the unit tests.

pysb/importers/sbml.py
```diff
@@ -15,7 +15,10 @@
 from pysb.logging import get_logger, EXTENDED_DEBUG
 
 BIOMODELS_REGEX = re.compile(r'(BIOMD|MODEL)[0-9]{10}')
-BIOMODELS_URL = 'http://www.ebi.ac.uk/biomodels-main/download?mid={}'
+BIOMODELS_URLS = {
+    'EBI': 'http://www.ebi.ac.uk/biomodels-main/download?mid={}',
```
jmuhlich: Could the keys be all lowercase, `'ebi'` and `'caltech'`? That seems more pythonic to me.

PR #328, opened 2017-12-04
Dangling bond detection for ReactionPattern

alubbock:
A dangling bond is when a specified bond has no binding partner, e.g.

```python
A(a=None) % A(a=1)
```

Patterns with dangling bonds are disallowed in Kappa and pending deprecation in BioNetGen. This PR adds dangling bond detection when `ReactionPattern` patterns, allowing users to detect these mistakes early (before network generation).

### DIFF
--- a/pysb/pattern.py
+++ b/pysb/pattern.py
@@ -80,17 +80,13 @@ def get_bonds_in_pattern(pat):
     Examples
     --------
 
-    These examples convert the set to a list for testing purposes only (sets
-    print differently on Python 2 vs Python 3). This is not necessary for
-    end users.
-
     >>> A = Monomer('A', ['b1', 'b2'], _export=False)
-    >>> list(get_bonds_in_pattern(A(b1=None, b2=None)))
-    []
-    >>> list(get_bonds_in_pattern(A(b1=1) % A(b2=1)))
-    [1]
-    >>> list(get_bonds_in_pattern(A(b1=1) % A(b1=2, b2=1) % A(b1=2)))
-    [1, 2]
+    >>> get_bonds_in_pattern(A(b1=None, b2=None)) == set()
+    True
+    >>> get_bonds_in_pattern(A(b1=1) % A(b2=1)) == {1}
+    True
+    >>> get_bonds_in_pattern(A(b1=1) % A(b1=2, b2=1) % A(b1=2)) == {1, 2}
+    True
     """
     return set(get_half_bonds_in_pattern(pat))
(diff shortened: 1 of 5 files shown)


pysb/pattern.py
```diff
@@ -1,25 +1,27 @@
...
-    ReactionPattern, ANY, as_complex_pattern
+    ReactionPattern, ANY, as_complex_pattern, DanglingBondError
 import networkx as nx
 from networkx.algorithms.isomorphism.vf2userfunc import GraphMatcher
 from networkx.algorithms.isomorphism import categorical_node_match
 import numpy as np
 
 
-def get_bonds_in_pattern(pat):
+def get_bonds_in_pattern(pat, as_set=True):
```
jmuhlich: This new interface seems unwieldy. A suggestion: Rename this function to `get_half_bonds_in_pattern(pat)` and have it always return the `list`. Then implement `get_bonds_in_pattern` as:
```python
def get_bonds_in_pattern(pat):
    return set(get_half_bonds_in_pattern(pat))
```
Then `check_dangling_bonds` would call `get_half_bonds_in_pattern`.

pysb/pattern.py
```diff
@@ -63,7 +57,42 @@ def _get_bonds_in_monomer_pattern(mp):
...
+
+    Returns
+    -------
+    set
+        Bond numbers used in the supplied pattern
+
+    Examples
+    --------
+
+    These examples convert the set to a list for testing purposes only (sets
```
jmuhlich: Oof, this is unfortunate. What about writing the examples slightly differently to sidestep the issue? Maybe with equality tests:
```python
>>> get_bonds_in_pattern(A(b1=None, b2=None)) == set()
True
>>> get_bonds_in_pattern(A(b1=1) % A(b2=1)) == {1}
True
...
```

alubbock:
Updated doctests as requested.

bgyori:
Clarification question here: is this also meant to raise an error for a ReactionPattern like
```
A(b=1) + B(a=1)
```
which is technically not "dangling"?

lh64:
Those would be considered dangling bonds. Both reactants in that case have a bond ‘1’ that doesn’t have a binding partner. So that should result in an error.

alubbock:
The pattern `A(b=1) + B(a=1)` does indeed raise an error now in PySB master, as it should -- I agree with Leonard's logic, because bonds cannot span across separate `ComplexPattern`s.

PR #329, opened 2017-12-04
Add networkx to sphinx mocks

alubbock:

PR #331, opened 2017-12-15
Add Frequently Asked Questions to docs

alubbock:
I've added some initial FAQs to the documentation, which we can build off of over time.

Based off of PR #329.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -13,24 +13,14 @@ General
 
 * Do you support Python 3?
 
-    **Yes, with one caveat, which will be fixed in the next release**. PySB
-    currently uses the `<>` operator for reversible rules, e.g.::
+    **Yes**. The current release of PySB supports Python 3.6 and
+    Python 2.7. Other Python 3.x releases are not explicitly tested.
 
-       Rule('bindA', A(b=None) + A(b=None) <> A(b=1) % A(b=1), kf, kr)
-
-    PySB "overloads" this operator from base Python, but this operator
-    (originally "not equals", a synonym for `!=`) was removed in Python 3.
-    Thus, to use Python 3 with the current PySB release, you will have to
-    write the above rule as the following::
-
-        Rule('bindA', A(b=None) + A(b=None) != A(b=1) % A(b=1), kf, kr)
-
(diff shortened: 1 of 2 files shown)


doc/faq.rst
```diff
@@ -0,0 +1,99 @@
...
+`Gitter channel <https://gitter.im/pysb/pysb>`_.
+For more general Python related questions, we recommend `Stack
+Overflow <https://www.stackoverflow.com>`_.
+
+General
+=======
+
+* Do you support Python 3?
+
+    **Yes, with one caveat, which will be fixed in the next release**. PySB
```
jmuhlich: Now that #309 has been merged, this whole answer can basically be replaced with "YES!"

alubbock: Updated FAQ on Python 3 support. I left a note in place about the previous reversible rule syntax and that it needs to be replaced for Python 3 support.

doc/faq.rst
```diff
@@ -0,0 +1,89 @@
...
+For more general Python related questions, we recommend `Stack
+Overflow <https://www.stackoverflow.com>`_.
+
+General
+=======
+
+* Do you support Python 3?
+
+    **Yes**. The current release of PySB supports Python 3.6 and
+    Python 2.7. Other Python 3.x releases but are not explicitly tested.
```
jmuhlich: I think there's an extraneous "but" in this sentence.

doc/faq.rst
```diff
@@ -0,0 +1,89 @@
...
+    value will come from the first condition which evaluates to True. Thus,
+    for the Expression to always have a value, the last condition should
+    default to True.
+
+Simulation
+==========
+
+* How can I speed up my `ScipyOdeSimulator` simulation?
+
+    **Check the weave library is installed.** `weave` is a Python library
```
jmuhlich: Since #320 is imminent, maybe we should suggest installing cython instead?

jmuhlich: Nevermind, as you said in the other PR, weave still seems faster. We'll resolve this later.

doc/faq.rst
alubbock: Fixed.

PR #332, opened 2017-12-24
Fix cupsoda simulator error on Tyson model

alubbock:
CupSodaSimulator was erroring on the models with rates consisting
of a single Parameter or Expression. Calling .args on these returns
a string, not a pysb.Component object, which fails when we try to
multiply together to get the rate. This commit introduces a check
and workaround in the CupSodaSimulator.

I've also removed the pycuda.autoinit import, which triggers GPU
initialization when pysb.simulator is imported, even if a
CupSodaSimulator is not instantiated. Instead, pycuda is initialized
on demand within the n_blocks function.

### DIFF
--- a/pysb/simulator/cupsoda.py
+++ b/pysb/simulator/cupsoda.py
@@ -13,19 +13,16 @@ import logging
 from pysb.logging import EXTENDED_DEBUG
 import shutil
 from pysb.pathfinder import get_path
+import sympy
 
 try:
     import pandas as pd
 except ImportError:
     pd = None
 try:
-    import pycuda.autoinit
     import pycuda.driver as cuda
-
-    use_pycuda = True
 except ImportError:
-    use_pycuda = False
-    pass
+    cuda = None
 
 
 class CupSodaSimulator(Simulator):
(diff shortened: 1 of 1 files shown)


alubbock:
@JamesPino Are you saying the `n_blocks` default doesn't work very well in general? If making pycuda a hard requirement will improve efficiency through better choice of `n_blocks`, then I'd personally be fine with that. Just need to make sure that:

1. `CupSodaSimulator` will still *import* without pycuda installed, even if it doesn't *run* without pycuda.
2. `cuda.init()` is only called within `CupSodaSimulator.__init__()`, to avoid GPU initialization at import time (the behaviour before this PR).

Feel free to push a commit to alubbock/pysb cupsodafix branch if you like, to add the change to this PR.

JamesPino:
Sorry for the confusion. For some reason I was thinking pycuda was required (for the gpu ssa code it is). This does as you intend.

alubbock:
No worries, will leave as is then.

pysb/simulator/cupsoda.py
```diff
@@ -462,8 +463,11 @@ def _get_cmatrix(self):
         par_vals = self.param_values[:, rate_mask]
         rate_order = []
         for rxn in self._model.reactions:
-            rate_args.append([arg for arg in rxn['rate'].args if
-                              not re.match("_*s", str(arg))])
+            if isinstance(rxn['rate'], pysb.Component):
```
jmuhlich: I think this if/else could be eliminated by using more of the sympy API:
```python
rate_args.append([arg for arg in rxn['rate'].atoms(sympy.Symbol) if not arg.name.startswith('__s')])
```
(I also made the assumption that `.startswith('__s')` is what you are actually after here to match the species symbols, rather than that `_*s` regex)

alubbock: That works -- thanks.

Issue #333, opened 2018-01-12
symbolic functions wrongly exported to kappa


glucksfall:
Hello,

I declared the following in PySB `Expression('term', -1 + sympy.sqrt(1 + 8 * RT / KD))` and it is exported as `%var: 'term' -1 + sqrt(1 + 8*'RT'*'KD'^(-1))` while it should be `%var: 'term' -1 + [sqrt](1 + 8*'RT'*'KD'^(-1))`. Issue tested with kasim3 and kasim4 (checkout d69ddb6)

Best regards

alubbock:
Hi @glucksfall, I've posted a fix for this in pull request #334, which also includes Kappa 4 support. When that gets merged, your issue should be resolved.

Issue #288, opened 2017-07-05
Expression in observables


glucksfall:
Hello,

I have a doubt how to write the following `pysb.Observable('PGA2PGA3', pga2() + pga3())` to produce `%obs: 'PGA2PGA3' (|pga2()| + |pga3()|)` kappa code

Thank you for any suggestion

bgyori:
I was looking for something similar recently. It looks like these lines: https://github.com/pysb/pysb/blob/85625b234bfb86fe91db859bd71fbc05fcad5c30/pysb/generator/kappa.py#L118-L127
currently produce a comma separated list of the terms of the reaction pattern, whereas the arithmetic expression you give above would actually work with KaSim. Should I try implementing this in the KappaGenerator, or is there some reason we shouldn't pursue this?

jmuhlich:
What does `|pga2()| + |pga3()|` mean for the Kappa observable? I can't find any documentation on that syntax.

bgyori:
The preferred representation for amounts is now to put |-s around the monomer, and you can define algebraic expressions of these to observe. Relevant documentation in this section: http://dev.executableknowledge.org/docs/KaSim-manual-master/KaSim_manual.htm#x1-320003.5

PR #334, opened 2018-01-13
Kappa 4 support, expression improvements

alubbock:
This PR introduces support for Kappa 4, which is enabled by default.
To use Kappa 3 syntax, use the `pre4syntax=False` argument to
`run_simulation` or `run_static_analysis`. Note that Kappa now
requires agents on the reactant side to correspond to agents on the
product side of a rule (i.e. order is preserved). The "dot" agent
is implemented in PySB as `KappaDot`, and is necessary in
[certain situations documented in the Kappa Manual](https://github.com/Kappa-Dev/KaSim/blob/master/CHANGES.md#agent-order-in-rules-must-be-maintained).

For example, the following Kappa rule:

    // Agent A spawns a copy of agent B
    A(),. -> A(),B() @ 'r'

would be implemented in PySB as:

    Rule('spawnB', A() + KappaDot() >> A() + B(), r)

This PR also improves Expression support for Kappa, implementing
`log`, `exp`, `min`, `max`, `sin`, `cos`, `tan`, `mod` and `sqrt`
functions, if expressions via sympy's `Piecewise`, and the `pi`
symbol.

Fixes: #333
Fixes: #288

### DIFF
--- a/pysb/generator/kappa.py
+++ b/pysb/generator/kappa.py
@@ -1,7 +1,8 @@
 import pysb
 import sympy
-from re import sub
 import warnings
+from sympy.printing import StrPrinter
+from sympy.core import S
 
 # Alias basestring under Python 3 for forwards compatibility
 try:
@@ -12,10 +13,12 @@ except NameError:
 class KappaGenerator(object):
 
     # Dialect can be either 'complx' or 'kasim' (default)
-    def __init__(self, model, dialect='kasim', _warn_no_ic=True):
+    def __init__(self, model, dialect='kasim', pre4syntax=False,
+                 _warn_no_ic=True):
         self.model = model
         self.__content = None
         self.dialect = dialect
+        self.pre4syntax = pre4syntax
         self._warn_no_ic = _warn_no_ic
 
     def get_content(self):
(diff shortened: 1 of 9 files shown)


pysb/generator/kappa.py
```diff
@@ -216,22 +271,82 @@ def format_site_condition(site, state):
...
+            if -expr.exp is S.Half and not rational:
+                # Note: Don't test "expr.exp == -S.Half" here, because that
+                # will match -0.5, which we don't want.
+                return "1/[sqrt](%s)" % self._print(expr.base)
+
+        return super(KappaPrinter, self)._print_Pow(expr, rational)\
+            .replace('**', '^')
+
+    def _print_Pi(self, expr):
+        return '[pi]'
```
jmuhlich: I guess `pi` is special-cased in Kappa? There are a few other built-in constants like that in sympy but maybe the only useful one in this context is `E`. Is that special cased in Kappa? If not, it could be implemented here anyway as `return str(float(expr))`.

pysb/generator/kappa.py
```diff
@@ -12,10 +13,12 @@
 class KappaGenerator(object):
 
     # Dialect can be either 'complx' or 'kasim' (default)
-    def __init__(self, model, dialect='kasim', _warn_no_ic=True):
+    def __init__(self, model, dialect='kasim', pre4syntax=False,
```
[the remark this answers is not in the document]
alubbock: Not averse to dropping Kappa 3 support, since it doesn't lose us any features and upgrading Kappa is straightforward. Done.

pysb/generator/kappa.py
alubbock: Yes, `[pi]` is lower case in Kappa. It appears `E` works as `[E]`, but is undocumented. I implemented it as such.

pysb/kappa.py
```diff
@@ -443,3 +459,29 @@ def _parse_kasim_outfile(out_filename):
         raise Exception("problem parsing KaSim outfile: " + str(e))
 
     return recarr
+
+
+class KappaDot(pysb.MonomerPattern):
```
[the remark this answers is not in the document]
alubbock: Fixed `KappaDot()` BNG compatibility by exporting it as the BNG null species `0` (i.e. it gets ignored). Also implemented `KappaDot.__repr__()` as requested.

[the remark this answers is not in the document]
jmuhlich: Are Kappa's "dot" and BNG's "null" equivalent? Should we just have a PySB construct that maps to both? `None` is attractive...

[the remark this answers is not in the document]
jmuhlich: And if `None` doesn't work out we can use `pysb.core.Keyword` to define `NULL` or something like that.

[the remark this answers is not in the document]
alubbock: Will give `None` a try.

pysb/tests/test_kappa.py
```diff
@@ -197,13 +198,16 @@ def test_influence_map_kasa():
 def test_unicode_strs():
     Monomer(u'A', [u'b'], {u'b':[u'y', u'n']})
     Monomer(u'B')
-    Rule(u'rule1', A(b=u'y') + KappaDot() >> KappaDot() + B(),
+    Rule(u'rule1', A(b=u'y') + None >> None + B(),
```
jmuhlich: This is overloading an unrelated test. The `None` testing ought to live in its own test case.

alubbock: Done.

pysb/tests/test_kappa.py
```diff
@@ -183,13 +198,28 @@ def test_influence_map_kasa():
 def test_unicode_strs():
     Monomer(u'A', [u'b'], {u'b':[u'y', u'n']})
     Monomer(u'B')
-    Rule(u'rule1', A(b=u'y') >> B(), Parameter(u'k', 1))
+    Rule(u'rule1', A(b=u'y') + None >> None + B(),
```
jmuhlich: This change should be reverted so it's back to just testing the unicode thing.

alubbock: The `None`s are now required with Kappa 4 for transforming one agent into another, otherwise the error `Left hand side/right hand side agent mismatch` appears. I changed the reaction pattern in this test to a state change to avoid this.

jmuhlich: Of course, good fix.

webermarcolivier:
I want to define a reaction that synthetizes two products. In Kappa 4.0, we can define such a reaction with syntax:
`'synthesize_A_and_B' M(),.,. -> M(),A(),B() @ 'k'`

However, same reaction in pysb is producing only one ghost agent (dot) instead of two. Minimal example:
```python
from pysb import *
from pysb import kappa
from pysb.export import export
from pysb.kappa import run_simulation

Model()
Monomer('A')
Monomer('B')
Monomer('M')
Parameter('k', 3.0)
Rule('synthesize_A_and_B', M() + None + None >> M() + A() + B(), k)

kappaModel = export(model, 'kappa')
with open('model.kappa', 'w') as f:
    f.write(kappaModel)

run_simulation(model, time=10, points=10, verbose=True)
```
Even when forcing add operator with parenthesis `(M() + None) + None` to prevent `None + None` to be evaluated first.
Is this an inherent limitations in pysb syntax?

PR #335, opened 2018-01-17
networkx/pydot support for reading .dot files

alubbock:
* pygraphviz support is now optional (tests complete without it):
pydot is now an alternative for reading .dot (graphviz) files. 
pygraphviz support on Windows on recent versions of Python is
patchy at best - pydot supports more platforms and is easier to
install.
* When dot file reading is required (kappa interface), the results are
now returned in the (more flexible and maintained) networkx format.
The dot file itself can be read using pydot or pygraphviz (both are
tried by default) - see new pysb.util.read_dot() function. Networkx
can export back to pygraphviz if required, as updated docs show.
* Move Python 3 continuous integration on both travis and 
appveyor to Python 3.6, to standardise across platforms
* Update doctests to numpy 1.14 standard (no leading whitespace
inside bracket when printing arrays)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 dist: trusty
 python:
 - '2.7'
-- '3.5'
+- '3.6'
 addons:
     apt:
         sources:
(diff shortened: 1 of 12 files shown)


alubbock:
The only thing missing from networkx/pydot interface as far as I can see is a plot function like `view_pygraphviz`, which is in the networkx/pygraphviz interface. Potentially we could use (for demonstration purposes) the networkx plotting functions instead, but their layouts are sometimes not as good as graphviz.

jmuhlich:
Looks like `networkx.drawing.nx_pydot` does have a `graphviz_layout` function. So you can draw a graph with layout by `dot` like so:
```python
import networkx as nx
g = nx.drawing.nx_pydot.read_dot('doc/examples/mymodel.dot')
pos = nx.drawing.nx_pydot.graphviz_layout(g, prog='dot')
nx.draw_networkx(g, pos, with_labels=True)
```

alubbock:
OK, I'll update the examples and drop pygraphviz as an import option.

johnbachman:
I notice that this PR reads the influence map as a networkx MultiDiGraph, which allows multiple edges between nodes. Is this necessary for an influence map, or would a DiGraph (single edge between nodes) suffice? If the DiGraph would suffice, handling the MultiDiGraph is unnecessarily cumbersome as it involves an extra dictionary layer before edge information can be accessed.

johnbachman:
Never mind--we came up with a counter-example. MultiDiGraph it is...

PR #337, opened 2018-01-27
On demand observable trajectory calculation

alubbock:
Normally, observables are added to a model before running a simulation.
If a new observable is required, previously it was necessary to re-run the
simulation. This PR adds a new `observable` method to the
`SimulationResult` class, which can calculate observables trajectories 
on demand using any supplied MonomerPattern or ComplexPattern,
without re-running the simulation.

Toy example for EARM 1.0:

```python
from pysb.examples import earm_1_0
from pysb.simulator import ScipyOdeSimulator
from pysb import ANY
simres = ScipyOdeSimulator(earm_1_0.model, tspan=range(5)).run()

# Alias monomers, for convenience
m = earm_1_0.model.monomers

# Calculate observable: bound cSmac
simres.observable(m.cSmac(b=ANY))
```

The result is returned as a pandas Series.


pysb/simulator/base.py
```diff
@@ -914,6 +914,112 @@ def observables(self):
             raise ValueError('Model has no observables')
         return self._squeeze_output(self._yobs)
 
+    def observable(self, pattern):
+        """
+        Calculate a pattern's trajectories without adding to model
+
+        Normally, observables are added in advance to a model. If a new
+        observable is required, previously it was necessary to re-run the
```
jmuhlich: The docstring doesn't need to explain the previous behavior.

pysb/simulator/base.py
```diff
@@ -914,6 +914,112 @@ def observables(self):
...
+        Returns
+        -------
+        pandas.Series
+            Series containing the simulation trajectories for the specified
+            observable
+
+        Examples
+        --------
+
+        >>> from pysb.examples import earm_1_0
```
jmuhlich: I feel the example section should just succinctly demonstrate the extraction of a novel observable not defined in the model. Recapitulating an existing observable obscures the whole point of this cool feature!

PR #340, opened 2018-01-31
Filtering and updates for ComponentSets

alubbock:
Add two new functions to ComponentSets - filter() and apply()

filter() can be used to filter a ComponentSet by properties of its
Components - filtering by name, pattern, module name, value, or rule
rate is supported. Examples:

    model.rules.filter(name='^inhibit', pattern=m.cSmac())
    model.rules.filter(fwd_rate=lambda rate: rate < 1e-8)
    model.parameters.filter(value=1e4)

Since filter() returns a ComponentSet, further filter() operations
can be chained on the end.

apply() is used to update the components in a ComponentSet, e.g.
parameter values, names, rule rates. For example, to double the values
of parameters kc1, kc3 and kc5:

    model.parameters.filter(name='^kc(1|3|5)$').apply(
                value=lambda val: val * 2)

Components now also keep track of the modules in which they were
defined, which is useful for runtime filtering of large models defined
across multiple files or functions, e.g. with the EARM 2.0 model:

    len(model.components)  # 215 components
    len(model.components.filter(module_name='earm.albeck_modules'))  # 95 components


jmuhlich:
And an example of explicit use of built-in `filter` to use an operator other than `and`:
```python
ComponentSet(filter(lambda r: pf.name('^inhibit')(c) or pf.pattern(m.cSmac())(c), model.rules))
```
Ugh, an explicit comprehension might actually be cleaner:
```python
ComponentSet(r for r in model.rules if pf.name('^inhibit')(c) or pf.pattern(m.cSmac())(c))
```
Also, the API of the predicate functions could be either to return a function (as I've assumed so far) or to take a final optional arg and return the boolean result directly. In the latter case my example would look like this:
```python
ComponentSet(r for r in model.rules if pf.name('^inhibit', c) or pf.pattern(m.cSmac(), c))
```

alubbock:
I agree about `apply`, that can be dropped. Your example is cleaner.

I think your `filter` redesign could be simplified further, by implementing `__or__`, `__and__` etc. within the `FilterPredicate` class, which return a new composite `FilterPredicate`. That way, we'd have e.g.:

```python
from pysb.pattern import FilterPredicate as FP
model.parameters.filter(FP.name('^kc(1|3|5)$') & FP.value(min=1e4, max=1e6))
model.rules.filter(FP.name('^inhibit') | FP.pattern(m.cSmac()))
```

One advantage of this approach is that a complex set of predicates could be assigned to a variable and reused.

Agree the `FilterPredicate` class could have an optional second argument to work on a `ComponentSet` directly, perhaps additionally with their own `filter` method:

```python
inhibited_or_cmac = FP.name('^inhibit') | FP.pattern(m.cSmac())
inhibited_or_cmac.filter(list_of_rules)
```

jmuhlich:
I had considered the boolean operator overloading idea but thought it might add too much overhead to creating user-defined predicate functions. I think the individual filter functions wouldn't be methods on a single FilterPredicate class, rather they'd be subclasses of an abstract base. A subclass's constructor would accept the pattern, min/max, etc. and its `__call__` method would take a single argument (a Component) and return True/False.

We could make the `__and__`, `__or__` etc. methods in the abstract base accept any callable with a single positional argument as the "other" to make user-defined predicates easier to write. Then users could just write plain functions or lambdas. Also we could provide a decorator to transform a plain function into a FilterPredicate instance if it was needed alone without chaining off an actual subclass. Some examples:

```python
# Automatic "upgrade" of lambda
model.parameters.filter(pf.Name('^kc(1|3|5)$') & lambda c: c.value / 100 < 7)


# Explicit annotation
@pf.predicate
def odd(c):
    return isinstance(c, Parameter) and c.value % 2 == 1

model.parameters.filter(odd & lambda ...)


# Could also use annotation as a wrapper
def odd(c):
    ...

model.parameters.filter(predicate(odd) & ...)
```

alubbock:
Like it. I'll have a go at a draft implementation soon.

alubbock:
@jmuhlich I've updated the PR with a new draft of this feature. Here are the examples from the docstring:

```python
>>> from pysb.examples.earm_1_0 import model
>>> from pysb.pattern import Name, Pattern, Module, Function
>>> m = model.monomers

# Find parameters exactly equal to 10000:
>>> model.parameters.filter(lambda c: c.value == 1e4)  \
    # doctest:+NORMALIZE_WHITESPACE
ComponentSet([
 Parameter('pC3_0', 10000.0),
 Parameter('pC6_0', 10000.0),
])

# Find rules with a forward rate < 1e-8, using a custom function:
>>> model.rules.filter(lambda c: c.rate_forward.value < 1e-8) \
    # doctest: +NORMALIZE_WHITESPACE
ComponentSet([
 Rule('bind_pC3_Apop', Apop(b=None) + pC3(b=None) | Apop(b=1) %
        pC3(b=1), kf25, kr25),
])

# We can also use some built in predicates for more complex matching
# scenarios, including combining multiple predicates.

# Find rules with a name beginning with "inhibit" that contain cSmac:
>>> model.rules.filter(Name('^inhibit') & Pattern(m.cSmac())) \
    # doctest: +NORMALIZE_WHITESPACE
ComponentSet([
 Rule('inhibit_cSmac_by_XIAP', cSmac(b=None) + XIAP(b=None) |
        cSmac(b=1) % XIAP(b=1), kf28, kr28),
])

# Find rules with any form of Bax (i.e. Bax, aBax, mBax):
>>> model.rules.filter(Pattern(m.Bax) | Pattern(m.aBax) | \
        Pattern(m.MBax)) # doctest: +NORMALIZE_WHITESPACE
ComponentSet([
 Rule('bind_Bax_tBid', tBid(b=None) + Bax(b=None) |
      tBid(b=1) % Bax(b=1), kf12, kr12),
 Rule('produce_aBax_via_tBid', tBid(b=1) % Bax(b=1) >>
      tBid(b=None) + aBax(b=None), kc12),
 Rule('transloc_MBax_aBax', aBax(b=None) |
      MBax(b=None), kf13, kr13),
 Rule('inhibit_MBax_by_Bcl2', MBax(b=None) + Bcl2(b=None) |
      MBax(b=1) % Bcl2(b=1), kf14, kr14),
 Rule('dimerize_MBax_to_Bax2', MBax(b=None) + MBax(b=None) |
      Bax2(b=None), kf15, kr15),
 ])

# Count the number of parameter that don't start with kf (note the ~
# negation operator):
>>> len(model.parameters.filter(~Name('^kf')))
60

# Get components not defined in this module (file). In this case,
# everything is defined in one file, but for multi-file models this
# becomes more useful:
>>> model.components.filter(~Module('^pysb.examples.earm_1_0$'))
ComponentSet([
 ])

# Count the number of rules defined in the 'catalyze' function:
>>> len(model.rules.filter(Function('^catalyze$')))
24
```

pysb/pattern.py
```diff
@@ -1,17 +1,151 @@
...
+    def __call__(self, component):
+        return not self.pred(component)
+
+    def __repr__(self):
+        return '~{}'.format(self.pred.__repr__())
+
+
+class Name(FilterPredicate):
+    """
+    Predicate to filter a ComponentSet by regular expression name match
```
jmuhlich: Should mention that this is a **search** rather than a **match**. Something like this:
> Predicate to filter a ComponentSet by regular expression name search
>
> Note that this uses `re.search` which matches anywhere in the component name.
> Use `^` to explicitly anchor the match to the beginning.

pysb/pattern.py
```diff
@@ -1,17 +1,151 @@
...
+
+    def __or__(self, other):
+        return _OrPredicate(self, other)
+
+    def __invert__(self):
+        return _NotPredicate(self)
+
+
+class _AndPredicate(FilterPredicate):
+    """ Combine two predicates with AND. Used by FilterPredicate. """
```
jmuhlich: Suggest new text:
> Logical conjunction of two predicates.
>
> This class is not generally used directly -- it exists to support operator overloading
> in the base FilterPredicate class.

pysb/pattern.py
```diff
@@ -1,17 +1,151 @@
...
+
+    def __call__(self, component):
+        return self.pred1(component) and self.pred2(component)
+
+    def __repr__(self):
+        return '{} & {}'.format(self.pred1.__repr__(), self.pred2.__repr__())
+
+
+class _OrPredicate(FilterPredicate):
+    """ Combine two predicates with OR. Used by FilterPredicate. """
```
jmuhlich: Same as above, but starting with:
> Logical disjunction of two predicates.

pysb/pattern.py
```diff
@@ -1,17 +1,151 @@
...
+
+    def __call__(self, component):
+        return self.pred1(component) or self.pred2(component)
+
+    def __repr__(self):
+        return '{} | {}'.format(self.pred1.__repr__(), self.pred2.__repr__())
+
+
+class _NotPredicate(FilterPredicate):
+    """ Negate a predicate with NOT. Used by FilterPredicate"""
```
jmuhlich: Same as above, but starting with:
> Logical negation of a predicate.

pysb/pattern.py
```diff
@@ -1,17 +1,151 @@
...
+            raise ValueError('Cannot apply pattern to {}'.format(
+                             component.__class__.__name__))
+
+    def __repr__(self):
+        return '{}({})'.format(self.__class__.__name__, self.pattern)
+
+
+class Module(FilterPredicate):
+    """
+    Predicate to filter a ComponentSet by pattern
```
jmuhlich: Replace with:
> Predicate to filter a ComponentSet by module where components are defined

pysb/pattern.py
```diff
@@ -1,17 +1,151 @@
...
+    def __init__(self, regex):
+        self.regex = regex
+
+    def __call__(self, component):
+        return any(re.search(self.regex, m) for m in component._modules)
+
+
+class Function(FilterPredicate):
+    """
+    Predicate to filter a ComponentSet by function
```
jmuhlich: Replace with:
> Predicate to filter a ComponentSet by function where components are defined

PR #348, opened 2018-03-13
Forbid defining a state as None

alubbock:
A state, as opposed to a bond, should not be able to be defined as
None. Doing so causes an error in BNG: "reaction list cannot be read
because of errors". This commit adds a check for this condition
and a unit test.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -362,7 +362,7 @@ class MonomerPattern(object):
         invalid_sites = []
         for (site, state) in site_conditions.items():
             # pass through to next iteration if state type is ok
-            if state == None:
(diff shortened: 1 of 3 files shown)


pysb/core.py
```diff
@@ -362,7 +362,7 @@ def __init__(self, monomer, site_conditions, compartment):
         invalid_sites = []
         for (site, state) in site_conditions.items():
             # pass through to next iteration if state type is ok
-            if state == None:
+            if state is None and site not in monomer.site_states.keys():
```
jmuhlich: Written more simply as `... not in monomer.site_states`.

alubbock: Fixed

Issue #349, opened 2018-03-16
Expression doesn't verify its argument is a sympy.Expr


jmuhlich:
`Expression` requires that its `expr` field is a `sympy.Expr`. The constructor doesn't validate this. We should raise a ValueError if another type is passed. It might be OK to automatically convert ints and floats to the corresponding sympy types, but no other automatic conversion should be performed.

alubbock:
I would suggest no automatic conversion. ints and floats are better represented as Parameters, since some simulators don't support Expressions. This choice between Parameter or Expression can be handled within the importer machinery for the SBML/BNGL import cases.

PR #350, opened 2018-03-16
Expressions must be created from sympy.Expr

alubbock:
The BNGL and SBML importers were creating Expressions from ints and
floats, which are better represented as Parameters. This PR
performs this conversion within the importers where applicable.

Fixes: #349

### DIFF
--- a/pysb/importers/bngl.py
+++ b/pysb/importers/bngl.py
@@ -214,7 +214,10 @@ class BnglBuilder(Builder):
             except KeyError:
                 return self.model.expressions[rate_law_name]
         elif rl.get('type') == 'Function':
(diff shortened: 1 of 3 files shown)


pysb/importers/bngl.py
```diff
@@ -312,7 +315,10 @@ def _parse_expressions(self):
                                                           expr_text,
                                                           ex.message))
             expr_namespace[expr_name] = expr_val
-            self.expression(expr_name, expr_val)
+            if isinstance(expr_val, (float, int)):
```
jmuhlich: Consider `isinstance(expr_val, numbers.Number)`.

PR #352, opened 2018-03-17
Validate site and state names

alubbock:
This introduces validation checks when declaring monomers that
site names are of the form `^_*[a-zA-Z][a-zA-Z0-9_]*$`, and state
values are of the form `^_*[a-zA-Z0-9][a-zA-Z0-9_]*$`. Kappa doesn't
support numeric state values anymore, so we should consider a warning
or deprecation for those in the future.

This also makes site/state validation errors raise a `ValueError` instead
of non-specific `Exception`s.


jmuhlich:
This disallows site names like `_1`. Where do those cause problems? I would have thought the same regex used in validating `Component.name` would work for both site names and state values.

alubbock:
Kappa disallows names like `_1` - their regex for component names is `_*[a-zA-Z][a-zA-Z0-9_−+]*`. We can't use the same regex if we want to allow numerical state values (which are obviously not valid variable names in Python). We could use the same regex for component and site names, but a more permissive one for state values.

How about: `_*[a-zA-Z][a-zA-Z0-9_]*` for component and site names, and `_*[a-zA-Z0-9][a-zA-Z0-9_]*` for state values? We could catch Kappa unsupported state values and patch them on export (numerical, or underscore prefixed numerical). The question is whether to disallow component and site names that are unsupported by Kappa (underscore prefixed numerical) or patch them on export, too.

CC @lh64

jmuhlich:
I just asked for clarification on whether `_1` etc. should be valid Kappa at Kappa-dev/KaSim#512 . The discussion there seems to imply it should be valid, but as you've observed the code says otherwise. I think we should be able to allow `[_a-zA-Z][_a-zA-Z0-9]*` for components and sites (same as Python variable names and our current `Component.name` rule), and `[_a-zA-Z0-9]+` for states, with the Kappa exporter adding prefixes as needed. Or we can use the former rule for both to be consistent everywhere, and prefix any problematic BNG/SBML names (digits only) on import.

jmuhlich:
The Kappa devs agreed it was a mistake, they did intend to allow `_1`. The one thing they do disallow is a sole single underscore.

alubbock:
I think patching offending patterns on Kappa export seems best. Since the Kappa devs only just made the change, we'd have to depend on Kappa master until they release a new version, or patch to something else valid in the meantime (e.g. `42` becomes `__state42__`).

jmuhlich:
I agree we should remain compatible with Kappa 4.0rc1, which means patching both `'42'` and `'_42'` (and `'_'` itself, for completeness). When they release an update we can remove our fix for `'_42'`.

The question remains, what should we adopt as the restriction within PySB itself? I think allowing state values like `'42'` might actually be too confusing, so I'm leaning toward the python identifier rule.

jmuhlich:
The downside of adopting the python identifier rule for site states means some valid BNG models will need fixups on import and thus won't round-trip.

jmuhlich:
Any thoughts on using the Python identifier naming rules for both sites and state values?

alubbock:
Summary from the commit message:

Site names, state names and state values are now validated against the
regex `[_a-z][_a-z0-9]*\Z`

Imports from BNG or SBML with numeric state values are patched with
leading underscores (a log message is generated per monomer when this
occurs).

Exports to Kappa with invalid state values `_*[0-9]*$` are patched
with a leading `__state` (a log message is generated per monomer when
this occurs). This requires the Kappa `format_*` functions have access
to the model to identify which state values have been changed, so
they're now methods of `KappaGenerator`. I've left one line wrappers for
each of the original top level functions to maintain API compatibility.

PR #354, opened 2018-04-06
Enforce synthesis products to be concrete

alubbock:
This PR introduces a check at rule instantiation that every
product is concrete for synthesis rules.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1200,6 +1200,15 @@ class Rule(Component):
         self.move_connected = move_connected
         # TODO: ensure all numbered sites are referenced exactly twice within each of reactants and products
 
+        # Check synthesis products are concrete
+        if self.is_synth():
+            rp = self.reactant_pattern if self.is_reversible else \
+                self.product_pattern
+            for cp in rp.complex_patterns:
+                if not cp.is_concrete():
+                    raise ValueError('Product {} of synthesis rule {} is not '
+                                     'concrete'.format(cp, self.name))
+
     def is_synth(self):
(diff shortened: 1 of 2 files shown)


pysb/core.py
```diff
@@ -1200,6 +1200,15 @@ def __init__(self, name, rule_expression, rate_forward, rate_reverse=None,
         self.move_connected = move_connected
         # TODO: ensure all numbered sites are referenced exactly twice within each of reactants and products
 
+        # Check synthesis products are concrete
+        if self.is_synth():
+            for cp in (self.reactant_pattern.complex_patterns if
```
jmuhlich: Suggest breaking this up a bit and using `any`:
```python
rp = self.reactant_pattern if self.is_reversible else self.product_pattern
if any(not cp.is_concrete() for cp in rp.complex_patterns):
    raise ...
```
That first line is pushing it on length, but even if that gets wrapped onto two lines I think it's more readable. Might want to wrap it so `rp` could be expanded to `reaction_pattern`.

alubbock:
My intention with these kinds of changes is to catch errors as soon in the modeling process as possible. This increases efficiency (particularly with large models which might take time to export/generate a network) and helps newbies - BNG errors are more difficult to understand for PySB users, as the message refers to the BNG syntax and line numbers.

Catching other instances of non-concrete product synthesis would require a more sophisticated rule parser which we don't currently have in PySB itself. The exception would be Kappa 4 style rules where reactant/product patterns have a one to one match, which is a case that could reasonably be implemented.

jmuhlich:
Good, we're in full agreement!

Issue #355, opened 2018-04-11
None state should be allowed in rule patterns


bgyori:
With the recent #348, the `None` value for sites is not allowed in rules anymore. This is an issue when trying to model a site with states (like a phosphosite which has 2 states) that another monomer can bind to irrespective of the phosphorylation state of that site. The model below  is an example where this comes up:
```
from pysb import *

Model()

Monomer('MEK', ['kinase'])
Monomer('ERK', ['phospho'], {'phospho': ['u', 'p']})

Parameter('kf_bind', 1.0)
Parameter('kr_bind', 1.0)
Parameter('MEK_0', 10000.0)
Parameter('ERK_0', 10000.0)

Rule('MEK_phosphorylation_ERK_phospho_fwd', MEK(kinase=None) + ERK(phospho=None) >> MEK(kinase=1) % ERK(phospho=1), kf_bind)

Initial(MEK(kinase=None), MEK_0)
Initial(ERK(phospho='u'), ERK_0)
```

alubbock:
#348 contained an error, which is now fixed in `master` by #356. Sorry for the mistake!

PR #356, opened 2018-04-12
Allow state=None in rule patterns

alubbock:
None should be allowed in MonomerPatterns, to allow for models like
the following:

    Monomer('A', ['phospho'], {'phospho': ['u', 'p']})
    A(phospho=None) + A(phospho=None) >> A(phospho=1) % A(phospho=1)

#348 missed the mark somewhat... my intention was to disallow undefined
states in *species*, not in *patterns*. Such patterns are already detected 
as un-concrete, so they are forbidden in initials.

Thanks to @bgyori for spotting and reporting this.


alubbock:
Fixes #355

PR #361, opened 2018-05-30
Fix BNG export of "log" function

jmuhlich:
The BNG syntax for logarithm is "ln", not "log". This updates the BngPrinter to format it correctly.


alubbock:
It'd be cleaner to overload `_print_log` than `_print_Function`:

```python
def _print_log(self, expr):
    # BNG doesn't accept "log", only "ln".
    return 'ln' + "(%s)" % self.stringify(expr.args, ", ")
```

And here's a simple unit test that could be added:

```python
@with_model
def test_bng_log_expr():
    Monomer('A')
    Expression('expr1', sympy.log(100))
    Rule('Rule1', None >> A(), expr1)
    generate_equations(model)
```

Issue #362, opened 2018-05-30
BNG error message extraction fails in the face of an underlying Perl warning


jmuhlich:
Saw this today:
> BngInterfaceError: 
> Unescaped left brace in regex is deprecated here (and will be fatal in Perl 5.30), passed through in regex; marked by <-- HERE in m/^(Rule[0-9]{ <-- HERE .*})r$/ at /home/jmuhlich/apps/BioNetGen-2.3.0/Perl2/Visualization/ProcessGraph.pm line 302.
> ABORT: Reaction rule list could not be read because of errors.

No further error messages were visible in the pysb exception.

Exporting the model to bngl and running BNG2.pl directly gives the full error:
>Unescaped left brace in regex is deprecated here (and will be fatal in Perl 5.30), passed through in regex; marked by <-- HERE in m/^(Rule[0-9]{ <-- HERE .*})r$/ at /home/jmuhlich/apps/BioNetGen-2.3.0/Perl2/Visualization/ProcessGraph.pm line 302.
>BioNetGen version 2.3.0
>Reading from file test.bngl (level 0)
>Read 6 parameters.
>Read 3 molecule types.
>Read 3 observable(s).
>Read 3 species.
>ERROR: Molecule created in reaction rule: Component(s) x missing from molecule B()
>  at line 30
>ABORT: Reaction rule list could not be read because of errors.

RuleWorld/bionetgen@46dd2c7d4d19884fdf100174712b5e1711ca625c fixes the Perl warning in BNG but the larger problem in pysb remains.

alubbock:
The error will also be visible if you set the PySB log level to DEBUG. It appears this is unrelated to the perl warning, instead it's down to the way the BNG output is read as a stream during execution, before the BNG exit code is checked.

I'll add some stream parsing to check for errors and warnings as they occur.

PR #363, opened 2018-05-30
Capture BNG errors into PySB exception

alubbock:
BNG output is now parsed for the strings 'ERROR:' and 'WARNING:'.
When encountered, the PySB log level is upgraded to ERROR/WARNING
for all future log output during that execute() call.

This applies to the BngFileInterface only. BngConsole already has
error and warning detection.

Fixes: #362


jmuhlich:
Note that the iteration over `p.stdout` consumes the data from the stream, so `p_out` is always empty and doesn't contain any `ERROR: ...` text from BNG. This needs a little thinking on what content actually belongs in logging vs. the exception message, anyway. I can only find one use of `logger.error` in all of pysb, in the cupsoda simulator, and that case also raises the same text in an exception right away. Is ERROR-level logging really useful for us?

Proposal for consideration: We keep using `logger.debug` for the BNG output as before. If ERROR: appears then that line and the ones after it would be added to a list. If that list is non-empty or the returncode is non-zero, those error lines and `p_err` get joined with `\n` and used as the exception message. I'm not sure if WARNING: lines from BNG need any special treatment. If we could pull out individual messages I'd say `warnings.warn` them, but I am guessing that would be fragile due to the way BNG formats and line-wraps them.

alubbock:
> Is ERROR-level logging really useful for us?

It might be, for recoverable errors, but those cases seem rare. One hypothetical scenario is a parallel simulator running a large number of simulations and one of them fails - you probably don't want to generate an Exception and lose all the other results, but you definitely want to know about the failure.

> Proposal for consideration... <snip>

Your proposal is close to the behaviour in the PR, except that logging is used rather than an Exception. All remaining lines after an error or warning are logged as errors or warnings. It should be easy enough to implement the error processing as you suggest (it seems a good approach).

I've spoken with @lh64 and BNG hasn't fully standardized its format for warnings, so parsing would be tricky to do reliably. I'll make this PR focus on errors only and we can tackle warnings in a future PR.

pysb/tests/test_bng.py
```diff
@@ -180,6 +180,13 @@ def test_sympy_parameter_keyword():
     generate_equations(model)
 
 
+def test_log():
```
jmuhlich: The `@with_model` annotation got lost here, probably as part of a conflict resolution. I know because I just did the same thing on a local rebase that added a new test at the end. :)

jmuhlich: This looks like the only reason CI is failing on this.

PR #364, opened 2018-05-30
Add fixed species support

jmuhlich:
This adds a `fixed` boolean kwarg to `Initial` for declaring that a species should not be consumed or produced by any reactions. This maps to a leading `$` on a species entry in BNG. If there's a Kappa equivalent I would be happy to add that too, but right now the KappaGenerator raises an exception.

The implementation adds a new list of booleans to `Model`, `initial_conditions_fixed`, rather than modify the current `initial_conditions` tuple structure. I'm not totally happy with this, though, and am open to alternatives. The effect of a fixed species is simply that its row in the stoichiometric matrix is set to zero.

### DIFF
--- a/pysb/export/python.py
+++ b/pysb/export/python.py
@@ -14,8 +14,8 @@ The standalone Python code defines a class, ``Model``, with a method
 ``simulate`` that can be used to simulate the model.
 
 As shown in the code for the Robertson model below, the ``Model`` class defines
-the fields ``parameters``, ``observables``, and ``initial_conditions`` as lists
-of ``collections.namedtuple`` objects that allow access to the features of the
+the fields ``parameters``, ``observables``, and ``initials`` as lists of
+``collections.namedtuple`` objects that allow access to the features of the
 model.
 
 The ``simulate`` method has the following signature::
@@ -146,7 +146,7 @@ class PythonExporter(Exporter):
             'num_species': len(self.model.species),
             'num_params': len(self.model.parameters),
             'num_observables': len(self.model.observables),
-            'num_ics': len(self.model.initial_conditions),
+            'num_ics': len(self.model.initials),
             }
         output.write(pad(r"""
             def __init__(self):
@@ -160,7 +160,7 @@ class PythonExporter(Exporter):
                 self.sim_param_values = numpy.empty(%(num_params)d)
                 self.parameters = [None] * %(num_params)d
                 self.observables = [None] * %(num_observables)d
-                self.initial_conditions = [None] * %(num_ics)d
+                self.initials = [None] * %(num_ics)d
             """, 4) % init_data)
         for i, p in enumerate(self.model.parameters):
             p_data = (i, repr(p.name), p.value)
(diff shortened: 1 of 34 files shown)


alubbock:
My suggestion, assuming you don't want to break API compatibility:

* Turn `Initial` into a proper class, which stores the pattern, initial value parameter/expression, and fixed flag in its instances. That way, we can extend it with new attributes in the future without breaking API compatibility.
* Store a list of `Initials` in `Model.initials` (or use `@property` to achieve same)
* Turn `Model.initial_conditions()` into a compatibility shim/wrapper for `Model.initials` and perhaps mark it as pending deprecation.

CC @lh64 in case he has input from BNG point of view

jmuhlich:
Agreed. I had in fact thought of making an `Initial` class but somehow missed that we could store them in `initials` and shim `initial_condtions` -- I was focused on how to store them in `initial_conditions` without breaking anything. It's actually lucky that for some reason I used `initial_conditions` when initially writing this. `initials` will nicely parallel the other class/container pairs (Monomer/monomers, Observable/observables, etc).

pysb/export/python.py
```diff
@@ -188,7 +187,7 @@ def ode_rhs(self, t, y, p):
                 ydot = self.ydot
                 weave.inline(r'''%s''', ['ydot', 't', 'y', 'p'])
                 return ydot
-            """, 8) % (pad('\n' + code_eqs, 16) + ' ' * 16))
+            """, 8) % ('\n' + pad(code_eqs, 16) + ' ' * 16))
```
jmuhlich: This just corrects some extraneous trailing whitespace on the first line of the output. I only noticed it because I re-generated robertson_standalone.py and git diff highlighted the trailing whitespace.

pysb/export/sbml.py
```diff
@@ -170,17 +169,20 @@ def convert(self, level=(3, 2)):
...
-            if isinstance(param, pysb.Expression):
+        fixed_species_idx = set()
+        for ic in self.model.initials:
+            sp_idx = self.model.get_species_index(ic.pattern)
+            if isinstance(ic.value, pysb.Expression):
                 ia = smodel.createInitialAssignment()
                 _check(ia)
                 _check(ia.setSymbol('__s{}'.format(sp_idx)))
-                init_mathml = self._sympy_to_sbmlast(sympify(param.name))
+                init_mathml = self._sympy_to_sbmlast(ic.value)
```
jmuhlich: sympify wasn't even needed here, the param object *is* a `sympy.Symbol` already.

jmuhlich:
Thanks again @alubbock for the gentle push to do the right thing here. :)  It's ready for another look now.

jmuhlich:
I'd rather have something on `Parameter` and `Expression` directly rather than provide this feature only in the context of `Initial`, since it could be helpful in other contexts. Also we probably don't want to facilitate a "transparent `.value`" concept TOO much, since it only operates on the nominal parameter values which are of limited utility in any serious application.

alubbock:
OK, let's address #400 separately, then.

Issue #269, opened 2017-05-05
SBML exporter does not work with compartments


alubbock:
PySB's SBML exporter currently outputs one compartment called `default`. It should also create SBML compartments for any compartments defined in PySB.

PR #367, opened 2018-06-08
LibSBML model exporter

alubbock:
This replaces the original SBML exporter with a libsbml-based one.
New features include compartment and expression support, and
conversion between SBML levels/versions (L3V2 by default).

Fixes: #269

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1615,9 +1615,19 @@ class Model(object):
         # intersect with original parameter list to retain ordering
         return self.parameters & cset
 
+    def parameters_expressions(self):
+        cset = ComponentSet()
+        for expr in self.expressions:
+            for sym in expr.expand_expr().free_symbols:
+                if sym in self.parameters:
+                    cset.add(sym)
+        # intersect with original parameter list to retain ordering
+        return self.parameters & cset
+
     def parameters_unused(self):
         """Return a ComponentSet of unused parameters."""
-        cset_used = self.parameters_rules() | self.parameters_initial_conditions() | self.parameters_compartments()
+        cset_used = (self.parameters_rules() | self.parameters_initial_conditions() |
+                     self.parameters_compartments() | self.parameters_expressions())
         return self.parameters - cset_used
 
 #     def expressions_constant(self):
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1616,6 +1616,7 @@ class Model(object):
         return self.parameters & cset
 
     def parameters_expressions(self):
+        """Return a ComponentSet of expression parameters."""
         cset = ComponentSet()
         for expr in self.expressions:
             for sym in expr.expand_expr().free_symbols:
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1616,7 +1616,7 @@ class Model(object):
         return self.parameters & cset
 
     def parameters_expressions(self):
-        """Return a ComponentSet of expression parameters."""
+        """Return a ComponentSet of the parameters used in expressions."""
         cset = ComponentSet()
         for expr in self.expressions:
             for sym in expr.expand_expr().free_symbols:
--- a/pysb/export/sbml.py
+++ b/pysb/export/sbml.py
@@ -8,6 +8,7 @@ import pysb
 import pysb.bng
 from pysb.export import Exporter
 from sympy.printing.mathml import MathMLPrinter
+from xml.dom.minidom import Document
 try:
     import libsbml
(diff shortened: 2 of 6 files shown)


pysb/export/sbml.py
```diff
@@ -37,51 +21,47 @@ def _print_Symbol(self, sym):
...
-        for a in annotations:
-            output += indent(format_single_annotation(a) + '\n', 20)
-        output += indent('    </rdf:Bag>\n</%s>\n' % qualifier, 12)
-    output += get_annotation_postamble()
-    return indent(output, 16)
+        return
+
+
+def _mathml_expr_call(expr, as_ast=False):
+    mathstr = '<apply><ci>{}</ci>'.format(expr.name)
```
jmuhlich: I'm nervous about building XML via string concatenation. `lxml.objectify` is great for succinctly building a DOM, although it would add a new dependency. It would look something like this:
```python
import lxml.objectify, lxml.etree

# Could construct these as globals
MATHML_NS = 'http://www.w3.org/1998/Math/MathML'
E = objectify.ElementMaker(annotate=False)
E_mathml = objectify.ElementMaker(
    namespace=MATHML_NS, nsmap={None: MATHML_NS}, annotate=False
)

# Function body
apply = E.apply()
apply.append(E.ci(expr.name))
# or if you like it on one line: apply = E.apply(E.ci(expr.name))
for sym in ...:
    apply.append(E.ci(sym.name ...))
if as_ast:
    math = E_mathml.math(apply)
    mathstr = lxml.etree.tostring(math)
    ...
else:
    return lxml.etree.tostring(apply)
```
Using the built-in `xml.dom.minidom` (as sympy's MathML printer does) probably wouldn't be that much more verbose, but I haven't worked with it much to say for sure.

pysb/export/sbml.py
```diff
@@ -90,84 +70,211 @@ class SbmlExporter(Exporter):
     Inherits from :py:class:`pysb.export.Exporter`, which implements
     basic functionality for all exporters.
     """
+    def __init__(self, *args, **kwargs):
+        super(SbmlExporter, self).__init__(*args, **kwargs)
+        if not libsbml:
```
jmuhlich: Suggest moving the libsbml check even before the super class init call to indicate it's absolutely required before we proceed.

pysb/export/sbml.py
```diff
@@ -90,84 +70,211 @@ class SbmlExporter(Exporter):
...
+            raise ImportError('The SbmlExporter requires the libsbml python package')
 
-    def export(self):
-        """Export the SBML for the PySB model associated with the exporter.
+    def _sympy_to_sbmlast(self, sympy_expr, wrap_lambda=False):
```
jmuhlich: This ought to be converted to DOM construction as well.

pysb/export/sbml.py
```diff
@@ -90,84 +70,211 @@ class SbmlExporter(Exporter):
...
+        """
+        Convert a sympy expression to the AST format used by libsbml
+
+        wrap_lambda indicates whether the expression should be converted into a mathml lambda function
+        """
+        mathml = print_mathml(sympy_expr)
```
jmuhlich: When converting to DOM building, this needs some thought. The sympy MathML printer returns a string which we'd have to run back through a parser to generate a DOM again. The XML strings should be short but it feels like a waste. The actual DOM object produced by the printer is behind the private method `Printer._print` so it feels a bit wrong to access that directly. And it looks like the `xml.dom.minidom` there isn't directly compatible with the DOM class in `lxml.etree`, so if we used `lxml` we'd have to round-trip anyway.

alubbock:
@jmuhlich Updated to make use of `xml.dom.minidom`. I've also added simulation tests for all example model SBML exports using [roadrunner](http://libroadrunner.org).

pysb/core.py
```diff
@@ -1615,9 +1615,19 @@ def parameters_compartments(self):
         # intersect with original parameter list to retain ordering
         return self.parameters & cset
 
+    def parameters_expressions(self):
+        cset = ComponentSet()
```
jmuhlich: Add docstring?

alubbock: Docstring added

jmuhlich: Sorry to nitpick, but can you change it to `Return a ComponentSet of the parameters used in expressions` (similar to `parameters_rules`)? The phrase "expression parameters" is sort of vague.

alubbock:
@jmuhlich Docstring updated as requested

PR #368, opened 2018-07-17
Test on Python 3.7 on Travis

alubbock:
Now that Python 3.7 has been released, we should add it to our tests.


alubbock:
Anaconda doesn't support Python 3.7 on Miniconda yet, but I'll keep an eye on it ([supported platforms listed here](https://www.appveyor.com/docs/build-environment/#miniconda))

PR #370, opened 2018-07-24
add observables to sbml export

FFroehlich:
As SBML does not provide any official support for "observables", I have here implemented the SBML export as AssignmentRule to a parameter.


FFroehlich:
great, thanks!

PR #374, opened 2018-09-11
Default module name when __name__ not set

alubbock:
The module/function name is captured by PySB when Components
are defined, but in some interpreters the `__name__` attribute
is not set for interactive sessions. To avoid an error, this
introduces a default value of `__unnamed__`.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -179,10 +179,7 @@ class Component(object):
         # We assume we're dealing with Component subclasses here
         frame = inspect.currentframe().f_back
         while frame is not None:
-            try:
-                mod_name = frame.f_globals['__name__']
-            except KeyError:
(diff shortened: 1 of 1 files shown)


pysb/core.py
```diff
@@ -179,7 +179,10 @@ def __init__(self, name, _export=True):
         # We assume we're dealing with Component subclasses here
         frame = inspect.currentframe().f_back
         while frame is not None:
-            mod_name = frame.f_globals['__name__']
+            try:
```
jmuhlich: Rather than the try/catch, suggest:

```python
mod_name = frame.f_globals.get('__name__', '__unnamed__')
```

alubbock: Yes, that's cleaner. Implemented.

PR #385, opened 2018-10-11
BngSimulator: respect cleanup flag in `__init__`

alubbock:
The cleanup flag in BngSimulator's `__init__()` was being ignored;
only the value specified to `run()` was used. The `__init__()`
cleanup flag is now treated as a default, which can be overridden
on a per simulation basis in `run()` if required.

### DIFF
--- a/pysb/simulator/bng.py
+++ b/pysb/simulator/bng.py
@@ -25,7 +25,7 @@ class BngSimulator(Simulator):
 
     def run(self, tspan=None, initials=None, param_values=None, n_runs=1,
             method='ssa', output_dir=None, output_file_basename=None,
-            cleanup=True, population_maps=None, **additional_args):
+            cleanup=None, population_maps=None, **additional_args):
         """
         Simulate a model using BioNetGen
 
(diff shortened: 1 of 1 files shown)


pysb/simulator/bng.py
```diff
@@ -152,7 +153,8 @@ def run(self, tspan=None, initials=None, param_values=None, n_runs=1,
                               verbose=verbose_bool,
                               output_dir=output_dir,
                               output_prefix=output_file_basename,
-                              cleanup=cleanup,
+                              cleanup=self.cleanup if cleanup is None
```
jmuhlich: The formatting here is a little unfortunate. Maybe just normalize the value of `cleanup` with `if cleanup is None: cleanup = self.cleanup` up with all the other parameter checking?

alubbock: Sounds good. Done.

pysb/simulator/bng.py
```diff
@@ -95,6 +95,9 @@ def run(self, tspan=None, initials=None, param_values=None, n_runs=1,
                                       _run_kwargs=locals()
                                       )
 
+        if cleanup is None:
+            clean = self.cleanup
```
jmuhlich: I think `clean` is a typo, should be `cleanup`.

alubbock: Gah, indeed! Fixed.

PR #387, opened 2018-10-25
Drug binding macro that mimics experimental settings when a perturbation is added after a reaction is initiated

ortega2247:
This macro generates the reversible binding reaction DRUG + SUBSTRATE | DRUG:SUBSTRATE that only gets triggered when the simulation reaches the time point t_action. The idea of this macro is to mimic experimental settings when a reaction is started and later on some kind of perturbation is added to the system. This type of reaction has been used fairly frequent in the Lopez lab and I thought it could be added as a macro.


alubbock:
I would suggest upgrading the Sphinx documentation notes to warnings, since it's important to be clear that using e.g. SSA with the macro will lead to incorrect results.

Otherwise, I think this looks good to merge.

ortega2247:
I just made the change from notes to warnings

PR #389, opened 2018-10-30
Use system PATH variable in pathfinder

alubbock:
This PR allows PySB to search paths defined in the system PATH variable for external executables. Paths in PATH have the lowest order of precedence (below specific environment variables, anaconda environments, and pathfinder's system default paths), so existing installations shouldn't be impacted.

### DIFF
--- a/pysb/pathfinder.py
+++ b/pysb/pathfinder.py
@@ -3,7 +3,7 @@ import sys
 import sysconfig
 
 # Set to False to not utilize the system PATH environment variable
-use_path = True
+use_path = 'PYSB_PATHFINDER_IGNORE_PATH' not in os.environ
 
 _path_config = {
     'bng': {
@@ -151,9 +151,8 @@ def get_path(prog_name):
 
     search_paths = path_conf['search_paths'][os.name]
     if use_path:
-        search_paths = list(search_paths) + [
-            path.strip('"') for path in
-            os.environ.get('PATH', '').split(os.pathsep)]
+        search_paths = list(search_paths) + os.environ.get('PATH', '').split(
+            os.pathsep)
 
     for search_path in search_paths:
         try:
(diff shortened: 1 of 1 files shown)


pysb/pathfinder.py
```diff
@@ -147,6 +150,11 @@ def get_path(prog_name):
                                      set_path.__name__))
 
     search_paths = path_conf['search_paths'][os.name]
+    if use_path:
+        search_paths = list(search_paths) + [
+            path.strip('"') for path in
```
jmuhlich: Why strip double-quote? Have you encountered this on some system? I think windows does support quoting directories with semicolons in them (it's a valid path character there), but this code won't handle that case anyway. Also, under POSIX systems it seems double-quote is not treated specially at all. (I tested in Bash / Linux to confirm this)

pysb/pathfinder.py
```diff
@@ -2,6 +2,9 @@
 import sys
 import sysconfig
 
+# Set to False to not utilize the system PATH environment variable
+use_path = True
```
jmuhlich: This doesn't feel like the right interface. What sort of situations would you expect people would want to turn this off? I think it would be user- or system-specific, not model- or script-specific. That would point to using an environment variable to control this rather than a Python global. Actually, perhaps both would be useful:
```python
use_path = 'PYSB_PATHFINDER_IGNORE_PATH' not in os.environ
```
This would initialize to True if the user didn't have that environment variable set, otherwise False. The user's Python code could then override it later.

pysb/pathfinder.py
alubbock: I think this was added in the original code preemptively. Agree with your feedback - I'll remove it unless/until a specific use case crops up later.

pysb/pathfinder.py
alubbock: Sounds good. Implemented as suggested.

PR #395, opened 2018-12-01
ScipyOdeSimulator multiprocessing capability

alubbock:
This PR uses the multiprocessing module to add parallel simulation
capability to the ScipyOdeSimulator. This only works when using
Cython as a compiler.

Multiprocessing is picky about what can be pickled and sent to
child processes, so the Cython code is sent as text for use with
inline(). Since inline() is called in `__init__()` with the same code,
it utilises cython's cache, and therefore the ODEs only have to be
compiled once. The code is a little bit duplicative, since we can't
pass instance methods through to multiprocessing's processes,
thefore the `rhs()` and `jac_fn()` methods must be redefined in
each process. Requires (and checks for) Python >=3.3.

### DIFF
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -29,6 +29,7 @@ import logging
 import itertools
 import contextlib
 import importlib
+import multiprocessing as mp
 
 
 class ScipyOdeSimulator(Simulator):
@@ -209,6 +210,7 @@ class ScipyOdeSimulator(Simulator):
             if self._compiler == 'cython':
                 if not Cython:
                     raise ImportError('Cython library is not installed')
+                self._code_eqs = code_eqs
 
                 def rhs(t, y, p):
                     # note that the evaluated code sets ydot as a side effect
@@ -270,6 +272,7 @@ class ScipyOdeSimulator(Simulator):
         # in case we want to do manipulations of the matrix later (e.g., to
         # put together the sensitivity matrix)
         jac_fn = None
+        self._jac_eqs = None
         if self._use_analytic_jacobian:
             species_symbols = [sympy.Symbol('__s%d' % i)
                                for i in range(len(self._model.species))]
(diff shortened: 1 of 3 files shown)


pysb/simulator/scipyode.py
```diff
@@ -592,3 +638,65 @@ def process(self, msg, kwargs):
     warn = logging.LoggerAdapter.info
     # Provide 'fatal' to match up with distutils log functions.
     fatal = logging.LoggerAdapter.critical
+
+def _lsoda_process(code_eqs, jac_eqs, num_species, num_odes, *args, **kwargs):
```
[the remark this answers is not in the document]
alubbock: Thanks, I wasn't familar with the `concurrent.futures` approach. I've implemented it as you suggested, and added a serial executor to avoid generating an extraneous process for the single processor version.

pysb/simulator/scipyode.py
```diff
@@ -209,40 +200,52 @@ def __init__(self, model, tspan=None, initials=None, param_values=None,
...
                     macro = arr_name.upper() + '1'
                     code_eqs = re.sub(r'\b%s\[(\d+)\]' % arr_name,
                                       '%s(\\1)' % macro, code_eqs)
 
-                def rhs(t, y, p):
-                    # note that the evaluated code sets ydot as a side effect
-                    weave_inline(code_eqs, ['ydot', 't', 'y', 'p'],
-                                 extra_compile_args=extra_compile_args)
-                    return ydot
+                    rhs = _get_rhs(self._compiler,
```
jmuhlich: This needs to be un-indented one level -- it's now inside the preceding loop!

pysb/simulator/scipyode.py
```diff
@@ -726,3 +676,17 @@ def jac_fn(t, y, p):
         trajectory[i:, :] = 'nan'
 
     return trajectory
+
+
+class DummyExecutor(Executor):
```
jmuhlich: Nice. I can't believe this isn't part of the standard `concurrent.futures` module. Nitpick: It should be called something like `SerialExecutor`, as "Dummy" would imply to me that it doesn't actually run anything.

pysb/simulator/scipyode.py
alubbock: Oops, good catch!

alubbock:
@jmuhlich Thanks, feedback implemented.

PR #402, opened 2019-01-09
Improve non-functional Cython error messages

alubbock:
On Windows, a lack of a functional compiler for Cython is often met with
the message "Unable to find vcvarsall.bat", which is cryptic for newbies.
This update adds a more descriptive warning complete with URL, logging,
and a consolidated "catch" condition.

### DIFF
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -403,35 +403,41 @@ class ScipyOdeSimulator(Simulator):
                                      extra_compile_args=extra_compile_args)
                     cls._use_inline = True
                 except (weave.build_tools.CompileError,
-                        distutils.errors.CompileError, ImportError):
-                    pass
-                except ValueError as e:
-                    if len(e.args) == 1 and \
-                                        e.args[0] == "Symbol table not found":
-                        get_logger(__name__).debug(
-                            "'ValueError: Symbol table not found' "
-                            "encountered; weave compiler is not functional")
-                    else:
+                        distutils.errors.CompileError,
+                        ImportError,
+                        ValueError) as e:
+                    if isinstance(e, ValueError) and (
+                            len(e.args) != 1 or
+                            e.args[0] != 'Symbol table not found'):
                         raise
+                    logger.warn('weave compiler appears non-functional. {}\n'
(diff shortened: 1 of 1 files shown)


pysb/simulator/scipyode.py
```diff
@@ -403,36 +403,42 @@ def _test_inline(cls):
...
-                    if len(e.args) == 1 and \
-                                        e.args[0] == "Symbol table not found":
-                        get_logger(__name__).debug(
-                            "'ValueError: Symbol table not found' "
-                            "encountered; weave compiler is not functional")
-                    else:
+                        distutils.errors.CompileError,
+                        ImportError,
+                        ValueError) as e:
+                    if isinstance(e, ValueError) and (
```
jmuhlich: This logic was already a little obscure and it's duplicated here and below. The immediate debug message previously made it fairly clear what was being checked, but with this change it's less apparent what's going on. Maybe this is a good time to refactor it. Could you refactor it into a well-named predicate function on `e` with an explanation of what it's looking for? Also I don't think rewriting the arg-checking term as a disjunction of negations made it any better... I do prefer the old form. Or maybe just `e.args == ('Symbol table not found',)` which actually covers the whole thing at once?

pysb/simulator/scipyode.py
```diff
@@ -403,36 +403,42 @@ def _test_inline(cls):
...
-                if len(e.args) == 1 and e.args[0] == "Symbol table not found":
-                    get_logger(__name__).debug(
-                        "'ValueError: Symbol table not found' "
-                        "encountered; Cython compiler is not functional")
-                else:
+            except (Cython.Compiler.Errors.CompileError,
+                    distutils.errors.DistutilsPlatformError,
+                    ValueError) as e:
+                if isinstance(e, distutils.errors.DistutilsPlatformError) and \
+                        str(e) != 'Unable to find vcvarsall.bat':
```
jmuhlich: Does this error not also occur under weave.inline? Can we unify all the C compiler breakage handling into one place?

alubbock:
I've refactored the logic into a separate function. Hopefully that's a little clearer now?

Issue #401, opened 2019-01-07
test_importers doesn't actually verify BNG simulation results


jmuhlich:
The `bngl_import_compare_simulations` function in `test_importers.py` looks like it should check the direct BNG simulation against a simulation of the same model round-tripped through pysb. However network generation hasn't been run on these models, so the line `for species in m.species` is problematic as `m.species` is always empty.

I think the we can compare the simulation result structures without needing a list of species, just from the structured dtype entries. (e.g. `yfull1.dtype.names`)

PR #406, opened 2019-01-10
Fix BioNetGen importer and related tests

alubbock:
* Ensure PySB imported model trajectories match BNG ones (#401)
* Employ reduced precision for checking models containing expressions
  with division. Sympy converts (x/y) to (x*y**(-1)), which results
  in lost precision where x and y are large.
* Fix import for models which contain (dynamic) expressions dependent
  on constant expressions. Previously, the latter were converted
  to generic sympy symbols during import.
* Respect `MoveConnected` keyword on BNGL import
* Respect observables' species compartments on BNGL import
* `match_once` flag shouldn't be set on observables which match species instead of molecules
* Preserve parameters in expressions, rather than convert them to
  numeric values on import.

Fixes: #401


pysb/importers/bngl.py
```diff
@@ -317,7 +317,9 @@ def _parse_rules(self):
             rule.rate_reverse = rev_rate
 
     def _parse_expressions(self):
-        expr_namespace = {p.name: p.value for p in self.model.parameters}
+        expr_namespace = {p.name: p for p in self.model.parameters}
```
jmuhlich: This can be drastically simplified since `ComponentSet` is dict-like and we aren't taking the `.value` of the parameters anymore:
```python
expr_namespace = dict(
    self.model.parameters | self.model.expressions_constant()
    | self.model.observables
)
```
(Some of that simplification could have been done previously, too, I guess) Depending on what's using it downstream, you probably don't even need to turn the combined ComponentSet into a dict.

alubbock:
Thanks for the feedback, suggestions have been implemented.

alubbock:
It seems that the new fixed species support brought to light a couple of issues importing BNG's `Motivating_example_cBNGL` model: the `MoveConnected` wasn't read in, species compartments in Observables were ignored, and the `match_once` flag shouldn't be set on observables which match species instead of molecules.

All the above issues have now been fixed in this PR.

pysb/tests/test_importers.py
```diff
@@ -34,15 +45,21 @@ def bngl_import_compare_simulations(bng_file, force=False,
...
-        print(yfull1[species])
-        print(yfull2[species])
-        print(numpy.allclose(yfull1[species], yfull2[species], atol=1e-8,
-                             rtol=1e-8))
-        assert numpy.allclose(yfull1[species], yfull2[species], atol=1e-8,
-                              rtol=1e-8)
+    assert yfull1.dtype.names == yfull2.dtype.names
+    for species in yfull1.dtype.names:
+        logger.debug(species)
+        logger.debug(species)
```
jmuhlich: Duplicated line.

pysb/tests/test_importers.py
```diff
@@ -34,15 +45,21 @@ def bngl_import_compare_simulations(bng_file, force=False,
...
+        logger.debug(yfull1[species])
+        logger.debug(yfull2[species])
+        logger.debug(numpy.allclose(yfull1[species], yfull2[species],
```
jmuhlich: Unsure if logging the result of `allclose` is useful, since the test passes/fails on that anyway.

pysb/tests/test_importers.py
```diff
@@ -9,9 +9,20 @@
 import mock
 import tempfile
 import shutil
+from pysb.logging import get_logger
+
+
+REDUCED_PRECISION = {
```
jmuhlich: Oh, I just meant to delete the part about the relative sizes of x and y impacting the precision of `x*y**-1`. The rest of that comment is important to keep to help understand why these models need reduced precision in the results comparison.

PR #408, opened 2019-01-24
Small changes to 'BngSimulator.run' to support optional arguments to

lh64:
BNG's 'generate_network' method. Added a new optional argument,
'netgen_args', to the 'BngSimulator.run' call signature. This is a
dictionary that gets passed to 'bngfile.action' when the
'generate_network' command is called.

### DIFF
--- a/pysb/simulator/bng.py
+++ b/pysb/simulator/bng.py
@@ -25,7 +25,8 @@ class BngSimulator(Simulator):
 
     def run(self, tspan=None, initials=None, param_values=None, n_runs=1,
             method='ssa', output_dir=None, output_file_basename=None,
-            cleanup=None, population_maps=None, **additional_args):
+            cleanup=None, population_maps=None, netgen_args={}, 
+            **additional_args):
         """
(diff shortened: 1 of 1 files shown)


lh64:
Ok, thanks.
1. I'll change `netgen_args={}` to `netgen_args=None`, as you suggest. 
2. I'll also add a test. The only issue here is that it's not actually possible to pass an argument that will cause an error. BNG doesn't check that your arguments are valid, so unrecognized arguments just end up getting ignored. I suppose we could fix that on the PySB side but it's true of all BNG methods (including `simulate`, etc.), so I'm not sure it makes sense. What do you think?
3. I think we can just pass kwargs to `generate_equations` for the other simulators. But those will need to be updated. `ScipyOdeSimulator`, for example, doesn't provide any way to pass kwargs over to `generate_equations`.

alubbock:
One other thing: `netgen_kwargs` is ignored if `method=='nf'`. Should probably raise a ValueError in that case, stating that the arguments are incompatible.

PR #409, opened 2019-02-02
cupSODA multi-GPU and chunk support

alubbock:
Add multi-GPU support and chunk support to cupSODA. Chunking allows the user to specify the maximum number of simulations per GPU per run. Larger numbers of simulations are split into chunks. This is useful for overcoming GPU RAM limitations.


pysb/simulator/cupsoda.py
```diff
@@ -250,58 +334,57 @@ def run(self, tspan=None, initials=None, param_values=None):
                                           param_values=param_values,
                                           _run_kwargs=[])
 
+        if isinstance(self.gpu, collections.Iterable):
```
jmuhlich: This kind of normalization probably belongs in `__init__` -- can you just apply this directly to `self.gpu` so it's always a list by the time the code gets to `run`?

PR #411, opened 2019-02-04
Use README.rst as long description in setup.py

alubbock:


alubbock:
@jmuhlich Good idea, implemented as requested.

PR #412, opened 2019-02-04
Local function support

alubbock:
Local functions are defined are mathematical rate law expressions which are evaluated separately for every molecular instance. This can greatly reduce the number of expressions required in a model.

See [Sneddon et al.](https://www.nature.com/articles/nmeth.1546) for further details and a motivating example.

The PR adds a new core construct, `pysb.core.Tag` and operator `@` for adding tags. Here is an example model automatically converted from [BioNetGen's localfunc.bngl example](https://github.com/RuleWorld/bionetgen/blob/master/bng2/Validate/localfunc.bngl):

```python
from pysb import Model, Monomer, Parameter, Expression, Compartment, Rule, Observable, Initial, MatchOnce, Annotation, Tag, ANY, WILD

Model()

Monomer('A', ['b', 'b', 'b'])
Monomer('B', ['a'])
Monomer('C')
Monomer('Trash')

Parameter('kp', 0.5)
Parameter('km', 0.1)
Parameter('k_synthC', 0.5)
Parameter('k_degrC', 0.5)
Parameter('Ab_b_b_0', 1.0)
Parameter('Ba_0', 3.0)
Parameter('C_0', 0.0)
Parameter('_Trash_0', 0.0)

Observable('Atot', A())
Observable('Btot', B())
Observable('Ctot', C())
Observable('AB0', A(b=MultiState(None, None, None)), match='species')
Observable('AB1', A(b=MultiState(1, None, None)) % B(a=1), match='species')
Observable('AB2', A(b=MultiState(1, 2, None)) % B(a=1) % B(a=2), match='species')
Observable('AB3', A(b=MultiState(1, 2, 3)) % B(a=1) % B(a=2) % B(a=3), match='species')
Observable('AB_motif', A(b=1) % B(a=1))

Tag('x')

Expression('f_synth', k_synthC*AB_motif(x)**2)
Expression('_rateLaw1', f_synth(x))

Rule('_R1', A() @ x >> A() @ x + C(), _rateLaw1)
Rule('_R2', A(b=None) + B(a=None) | A(b=1) % B(a=1), kp, km)
Rule('_R3', C() >> Trash(), k_degrC)

Initial(A(b=MultiState(None, None, None)), Ab_b_b_0)
Initial(B(a=None), Ba_0)
Initial(C(), C_0)
Initial(Trash(), _Trash_0, fixed=True)
```

Note that `x` has been defined as a tag, and the expression `f_synth` and rule `_R1` are dependent on that tag.


pysb/core.py
```diff
@@ -1551,6 +1636,39 @@ def __repr__(self):
...
+    def __call__(self, tag):
+        if not isinstance(tag, Tag):
+            raise ValueError('Expressions are only callable with a Tag '
+                             'instance, for use within local Expressions')
+
+        return sympy.Function(self.name)(tag)
+
+
+class Tag(Component, sympy.Symbol):
+    """ Tag for labelling MonomerPatterns and ComplexPatterns """
```
jmuhlich: Extraneous spaces just inside quotes.

pysb/core.py
```diff
@@ -988,6 +1006,8 @@ def __radd__(self, other):
             return NotImplemented
 
     def __mod__(self, other):
+        if self._tag is not None:
+            raise ValueError('Tag should be specified at the end of the complex')
```
jmuhlich: Is this actually a problem, or just bad style? If it's just a style issue, maybe a warning would be better.

pysb/importers/bngl.py
```diff
@@ -65,7 +66,8 @@ def _eval_in_model_env(self, expression):
 
         # Quick security check on the expression
         if re.match(r'^[\w\s()/+\-._*]*$', expression):
-            return parse_expr(expression, self._model_env)
+            return sympy.sympify(expression, locals=self._model_env,
```
jmuhlich: Why the change here?

pysb/importers/bngl.py
```diff
@@ -164,6 +178,7 @@ def _parse_monomers(self):
                             self._renamed_states[mon_name])
                         )
             try:
+                # self.monomer(mon_name, sites, dict(states))
```
jmuhlich: Delete commented line?

pysb/importers/bngl.py
```diff
@@ -358,24 +378,34 @@ def _parse_rules(self):
...
+                    tag = self.tag(tag_name)
+                    expr_namespace[tag_name] = tag
             expr_name = e.get('id')
             expr_text = e.find(_ns('{0}Expression')).text.replace('^', '**')
             expr_val = 0
             try:
                 expr_val = parse_expr(expr_text, local_dict=expr_namespace)
             except Exception as ex:
+                try:
+                    msg = ex.message
```
jmuhlich: `ex.message` doesn't exist in Python 3, the replacement is `str(ex)` which is backwards-compatible with Python 2. I guess the try/except would then need to be replaced with `if not msg:`.

pysb/importers/bngl.py
```diff
@@ -358,24 +378,34 @@ def _parse_rules(self):
     def _parse_expressions(self):
         expr_namespace = dict(
             self.model.parameters | self.model.expressions_constant()
-            | self.model.observables
```
jmuhlich: What happens with observables now?

pysb/core.py
```diff
@@ -649,6 +652,18 @@ def __pow__(self, other):
         else:
             return NotImplemented
 
+    def __matmul__(self, other):
```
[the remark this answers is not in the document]
alubbock: We're dropping Python 2.7 support at the end of the year (#414), so it doesn't seem worth adding a shim which would necessarily introduce new syntax for such a short period. 

Given that [Python <3.5 represented <5% of Python 3.x installation in 2018](https://www.jetbrains.com/research/python-developers-survey-2018/), supporting Python <3.5 doesn't seem worth it either.

Worth considering official Python version policy for PySB. Perhaps 3 most recent versions, i.e. 3.5-3.7 currently, moving to 3.6-3.8 when the latter is released. Thoughts?

pysb/bng.py
```diff
@@ -807,7 +875,7 @@ def _parse_species(model, line):
                     site_name, condition = ss, None
                 site_conditions[site_name].append(condition)
 
-        site_conditions = {k: v[0] if len(v) == 1 else tuple(v)
+        site_conditions = {k: v[0] if len(v) == 1 else pysb.core.MultiState(*v)
```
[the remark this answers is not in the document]
alubbock: No, multibonds use lists and will be a single entry in `v`. I've added tests for multiple bonds and MultiStates in #462, which also makes this change and hopefully clarifies things. Perhaps that simpler PR should get merged first?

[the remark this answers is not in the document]
jmuhlich: I see now, I forgot site_conditions is now a `Dict[str, list]` as part of the MultiState support. Looks like we just forgot to add model tests with MultiState, oops! :astonished:

pysb/core.py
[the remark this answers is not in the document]
jmuhlich: I guess `__matmul__` is fine as it is, then. I created #463 for the version support issue.

pysb/core.py
alubbock: It's a problem in some scenarios.

`A() % A() @ tag % B()` generates this error. Since we can't nest complexes or tag a subset of a species, the tag would have to get applied to the whole species `A() % A() % B()`, which might not be what the modeler intended.

The error would also come up if trying to bind two tagged species, e.g. `(A() % A() @ tag1) % (B() % B() @ tag2)`. Since we can't have two tags attached to the single combined species `A() % A() % B() % B()`, we'd have to choose a tag by some precedence system (or error out).

pysb/bng.py
```diff
@@ -770,6 +776,68 @@ def _parse_netfile(model, lines):
         pass
 
 
+def _parse_parameter(model, line):
+    index, pname, pval, hash, ptype = line.strip().split()
+    par_names = model.components.keys()
+    if pname not in par_names:
+        if ptype == 'Constant' and pname not in model._derived_parameters.keys():
+            try:
+                p = pysb.core.Parameter(pname, pval, _export=False)
```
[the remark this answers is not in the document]
alubbock: This issue doesn't seem to appear in the test suite, even when I tried fudging around with various forms. Anything with operators should, in theory, appear as a `ConstantExpression` instead of `Constant`, anyway. In the absence of a verifiable test case, I'll just remove this check.

pysb/bng.py
```diff
@@ -770,6 +776,68 @@ def _parse_netfile(model, lines):
...
+            except ValueError:
+                p = pysb.core.Parameter(pname,
+                                        eval(pval.replace('^', '**')),
+                                        _export=False)
+            model._derived_parameters.add(p)
+        elif ptype == 'ConstantExpression' and \
+                pname not in model._derived_expressions.keys():
+            pval = _fix_booleans(pval)
+            p = pysb.core.Expression(pname, sympy.sympify(pval),
```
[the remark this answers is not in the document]
alubbock: Yes, unification into a single BNGL expression parsing function is the way to go. ﻿﻿I've added such a function as `parse_bngl_expr` in pysb/bngl.py, based partly on your PR #456 and my comments on that thread.

pysb/bng.py
```diff
@@ -770,6 +776,68 @@ def _parse_netfile(model, lines):
         pass
 
 
+def _parse_parameter(model, line):
+    index, pname, pval, hash, ptype = line.strip().split()
```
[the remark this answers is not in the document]
alubbock: Fixed

pysb/bng.py
```diff
@@ -770,6 +776,68 @@ def _parse_netfile(model, lines):
...
+
+
+_RE_NUMBER = '\d+(?:\.\d+)?'
+_RE_EQUAL_INEQ = '{0}\s*([<>]=?|!=|==)\s*{0}'.format(_RE_NUMBER)
+_RE_AND = '({0})\s*&&\s*({0})'.format(_RE_NUMBER)
+_RE_OR = '({0})\s*\|\|\s*({0})'.format(_RE_NUMBER)
+_RE_NUMBER_PARENS = '\(\s*({0})\s*\)'.format(_RE_NUMBER)
+
+
+def _fix_booleans(pval):
```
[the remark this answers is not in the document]
alubbock: > could you explain the situations this handles?

Good question... it arose with one of the BNGL test models, but I can no longer reproduce it. Looks like it's basically down to the fact that sympy will handle this boolean-to-int casting as part of `lambdify`, but not with `subs`. Example (tested with sympy 1.4):

```python
import sympy

x = sympy.Symbol('x')
e = (x > 10) * 2

print('Should print "0" and "2"')
print(sympy.lambdify([x], e)(10))
print(sympy.lambdify([x], e)(11))

e.subs({x: 10})  # errors out
```

The Simulator class uses `lambdify`, and there aren't any current test failures without this code (perhaps the `parse_bngl_expr` refactoring helped?), I've removed it.

pysb/importers/bngl.py
alubbock: This is now altered to use `parse_bngl_expr`

pysb/importers/bngl.py
alubbock: Done

pysb/importers/bngl.py
alubbock: Replaced with `str(msg)` as suggested

pysb/importers/bngl.py
alubbock: This removal was probably a mistake. Reverted.

pysb/bng.py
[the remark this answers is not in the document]
jmuhlich: This feels good. :)

jmuhlich:
Thanks!

Issue #415, opened 2019-02-08
Simulations initials or param values supplied as pandas DataFrames may give incorrect results


alubbock:
The logic for processing initials and param_values currently handles data types other than dictionaries or numpy arrays by converting to numpy array. In the case of supplying a pandas DataFrame, this appears to work to the user, but leads to incorrect results where the DataFrame columns are in a different order to the model's initials/parameters.

PR #416, opened 2019-02-08
Support for simulation initials, params as DataFrames

alubbock:
Add support for supplying simulation initial conditions and parameter
values as pandas DataFrames. Previously this was handled by implict
conversion to numpy array, which might lose the column order (i.e.,
initials species or parameters get swapped).

Implicit conversion or simulation initials or param values from anything
other than a list is now disabled, meaning that users must now supply
a DataFrame, numpy array, or list for initials and param_values.

Fixes: #415

### DIFF
--- a/pysb/simulator/base.py
+++ b/pysb/simulator/base.py
@@ -334,6 +334,10 @@ class Simulator(object):
         if pd and isinstance(new_initials, pd.DataFrame):
             new_initials = new_initials.to_dict(orient='list')
 
+        # If new_initials is a list, convert to numpy array
+        if isinstance(new_initials, list):
+            new_initials = np.array(new_initials, copy=False)
+
         # Check if new_initials is a dict, and if so validate the keys
         # (ComplexPatterns)
         if isinstance(new_initials, dict):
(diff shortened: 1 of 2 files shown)


pysb/simulator/base.py
```diff
@@ -356,6 +360,11 @@ def _process_incoming_initials(self, new_initials):
                                      'values'.format(cplx_pat))
         else:
             if not isinstance(new_initials, np.ndarray):
+                if not isinstance(new_initials, list):
```
jmuhlich: Could this nested-not bit be refactored? Maybe normalize list->ndarray up top where DataFrame is normalized to dict, then this whole `else` can become `elif isinstance(new_initials, np.ndarray)` and the ValueError can be moved to a final `else` at the bottom.

pysb/simulator/base.py
```diff
@@ -469,6 +483,11 @@ def _process_incoming_params(self, new_params):
                                      "must be equal length")
         else:
             if not isinstance(new_params, np.ndarray):
+                if not isinstance(new_params, list):
```
jmuhlich: This would need the same treatment as `_process_incoming_initials` above.

alubbock:
Thanks, good suggestions again. Implemented.

PR #417, opened 2019-02-08
simplify initial condition export and preserve parameter dependency

FFroehlich:
If we directly write the values of `Parameters`/`Expression` as `InitialAmounts` or `InitialAssignment` we lose the dependency on the original entities. Also simplifies code a bit.


FFroehlich:
As there is no way of having a non-`Parameter/Expression` value for `Initials` it is probably not advisable to always encode them as symbolic entities. Yet, with the previous implementation it was impossible to create parameter dependent intial conditions from pysb. As a compromise, this PR implements the encoding of `Initial`s with `Expression` values as symbolic references in sbml instead of expanding them and writing the flat values. This allows creation of parametric initial conditions by creating an `Expression` for every initial condition parameter.

pysb/export/sbml.py
```diff
@@ -176,14 +177,16 @@ def convert(self, level=(3, 2)):
                 ia = smodel.createInitialAssignment()
                 _check(ia)
                 _check(ia.setSymbol('__s{}'.format(sp_idx)))
-                init_mathml = self._sympy_to_sbmlast(ic.value)
+                init_mathml = self._sympy_to_sbmlast(Symbol(ic.value.name))
                 _check(ia.setMath(init_mathml))
-                initial_concs[sp_idx] = None
             else:
                 initial_concs[sp_idx] = ic.value.value
                 if ic.fixed:
```
alubbock: Should remove this `if ic.fixed` block, since it's now duplicated below

FFroehlich:
Okay changed such that both `Expressions` and `Parameters` will be written as initial condition

alubbock:
Caught a couple of `.pytest_cache` files in the commit. The Appveyor build was also failing due to a separate issue with the stochkit-lite and BNG anaconda packages, which I've now fixed. Hopefully if you re-push with the `.pytest-cache` files removed, the new build will go through.

PR #418, opened 2019-02-08
fix exception when calling expand_obs() on Observable with no coeffs

FFroehlich:
`sympy.symbols(',')` throws a `ValueError`

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1352,8 +1352,7 @@ class Observable(Component, sympy.Symbol):
         """ Expand observables in terms of species and coefficients """
         return sympy.Add(*[a * b for a, b in zip(
             self.coefficients,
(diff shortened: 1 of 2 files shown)


FFroehlich:
Added the unit test. One thing I am a bit worried about is that you can now call `.expand_obs()` on observables before `pysb.bng.generate_equations` is called (as done in the unit test) and it will always return 0. I think in that setting `expand_obs()` should actually throw an exception and not return 0. Yet, I don't see a good way of checking whether `pysb.bng.generate_equations` has been called as checking for empty species/reactions will yield false positives for empty models.

alubbock:
I agree the behaviour is not ideal, but `model.reactions` and `model.species` both currently work similarly (they return empty lists before network generation). Ideally we'd track model state and (re-)generate the network on demand, but that's a PR for another day.

PR #419, opened 2019-02-11
make out of MonomerPattern.__repr__ deterministic

FFroehlich:
Especially noticable when exporting models to pysb_flat. Currently, ordering of sites in MonomerPatterns seems to be changing arbitrarily.


alubbock:
Can you provide an example which demonstrates non-deterministic ordering? `MonomerPattern.__repr__` prints sites in the order of `monomer.sites`, which is a list and hence should preserve the order that sites are specified when the monomer is instantiated. Are you seeing the site order vary in the `Monomer`s themselves?

FFroehlich:
Nondeterminancy in had indeed upstream origins, nothing related to pysb.

PR #426, opened 2019-03-22
Updating sensitivity analysis class to be general for any model parameters

JamesPino:
The class was built for initial concentrations. This update allows any parameters to be used. Also added the ability to pass a list of parameter names rather than requiring all from the model to be used.


JamesPino:
I believe I address all the comments. The travis test is failing for python 3.6 due to an hash error during libroadrunner installation. It passed in the [lolab-vu](https://travis-ci.org/LoLab-VU/pysb/builds/512746479) tests. I am unable to restart the tests for pysb/pysb.

pysb/tests/test_sensitivity_analysis.py
```diff
@@ -1,12 +1,13 @@
...
-    InitialsSensitivity
+    PairwiseSensitivity, InitialsSensitivity
 from pysb.examples.tyson_oscillator import model
 import numpy as np
 import numpy.testing as npt
 import os
 from pysb.simulator.scipyode import ScipyOdeSimulator
 import tempfile
 import shutil
+from nose.tools import *
```
alubbock: Should be `from nose.tools import raises`. I know there are wildcard imports in other tests - we should replace those in the future too. PEP8 says to avoid them.

pysb/tests/test_sensitivity_analysis.py
```diff
@@ -121,21 +125,91 @@ def test_create_png(self):
...
+            observable=self.observable,
+            solver=self.solver,
+            sens_type='all'
+        )
+        sens.run()
+        self.sens.create_plot_p_h_pprime(save_name='test4',
+                                         out_dir=None)
+        assert os.path.exists('test4_P_H_P_prime.png')
+
+    @raises(Exception)
```
alubbock: Use a more specific exception class. Otherwise, unrelated errors can end up getting missed. Likewise for the other tests in this file.

pysb/tools/sensitivity_analysis.py
```diff
@@ -112,72 +109,247 @@ class InitialsSensitivity(object):
...
     >>> sens.create_boxplot_and_heatplot() #doctest: +SKIP
+    >>> values_to_sample = [.9, 1.1]
+    >>> sens = PairwiseSensitivity(\
+            values_to_sample=values_to_sample,\
+            observable=observable,\
+            objective_function=obj_func_cell_cycle,\
+            solver=solver,\
+            sens_type='params'\
+        )
+    >>> print(sens.b_matrix)
```
alubbock: Not sure this huge matrix adds much to the example. I'd recommend removing it from the doctest, or changing it to print something shorter (like the matrix shape).

pysb/tools/sensitivity_analysis.py
```diff
@@ -112,72 +109,247 @@ class InitialsSensitivity(object):
...
-        if solver is None or not isinstance(solver,
-                                            pysb.simulator.base.Simulator):
-            raise(TypeError, "solver must be a pysb.simulator object")
         self._solver = solver
         self.objective_function = objective_function
         self.observable = observable
+        self._sens_type = sens_type
+        if self._sens_type not in ('params', 'initials', 'all'):
+            if sample_list is None:
+                raise Exception("Please provide 'sens_type' or 'sample_list'")
```
alubbock: Use `ValueError` here

pysb/tools/sensitivity_analysis.py
```diff
@@ -112,72 +109,247 @@ class InitialsSensitivity(object):
...
+        if sample_list is not None:
+            _valid_options = [i.name for i in self._model.parameters]
+            for i in sample_list:
+                if i not in _valid_options:
+                    raise Exception("{} not in model.parameters".format(i))
```
alubbock: Use `ValueError` here

pysb/tools/sensitivity_analysis.py
```diff
@@ -268,27 +440,15 @@ def run(self, save_name=None, out_dir=None):
...
         """ creates sample matrix, index of samples values, and shows bij """
         counter = -1
         sampled_values_index = set()
         bij_unique = dict()
         sampling_matrix = np.zeros(
-            (self._size_of_matrix, len(self._model.species)))
-        sampling_matrix[:, :] = self._original_initial_conditions
+            (self._size_of_matrix, len(self.orig_vals)))
+        sampling_matrix[:, :] = self.orig_vals
+        # print(sampling_matrix.shape)
```
alubbock: Remove commented line (unless there's a reason to keep it?)

alubbock:
Thanks, just some small details remaining that I can see:

* Add the `sens_type` parameter to the attributes list
* Ideally, attributes which are also parameters should appear in the parameters list. Maybe make the parameters version contain the description, then write `Identical to Parameters (see above)` in the attributes section, as is done in a few places in core.py
* Fix the Python 2.7 test on Windows (see Appveyor). Probably easiest to just remove the `print(sens.b_matrix.shape)` doctest (sorry, that was my suggestion - didn't realize it would break on Py27/Win!)

JamesPino:
Thanks for the feedback. I fixed that silly python2 on windows error that day, just forgot to push. I added the attributes and parameters. Some of them should have been private, as they should not have been used elsewhere. I really want to rewrite this entire code from scratch. It was written by a younger me who tried to be clever when I didn't need to be.

alubbock:
> I really want to rewrite this entire code from scratch
That's often the way! When we make the move to PySB 2 there will be other backwards-incompatible changes, so if you want to make those changes in a separate branch I'd say go ahead; it can get merged in when the time comes.

If I'm being nitpicky, there's a stray `O` in your docstring: `sens_type: {'params', 'initials', 'all'}        Type of sensitivity analysis to perform. O`. But otherwise, LGTM. Thanks again!

PR #427, opened 2019-03-22
Adding CUDA and openCL SSA

JamesPino:
Created two simulators for SSA simulations. One leverages CUDA through pycuda, the other openCL with pyopenCL. The pyopenCL allows the use of CPU and GPU. This has allowed substantial speedup compared to using BNG ssa. We are currently writing a manuscript demonstrating their utility.


pysb/examples/gpu_ssa/run_earm_gpu_ssa.py
```diff
@@ -0,0 +1,46 @@
+import logging
```
alubbock: Example model filenames should clarify which GPU SSA they're using. A short docstring at the top describing the example would be helpful, too.

pysb/examples/gpu_ssa/run_schlogl_gpu_ssa.py
```diff
@@ -0,0 +1,64 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+from pysb.bng import generate_network
+from pysb.examples.schlogl import model
+from pysb.integrate import odesolve
+from pysb.simulator.cuda_ssa import CUDASimulator
+
+generate_network(model, verbose=False, cleanup=False, )
```
alubbock: This should go in `main`

pysb/examples/gpu_ssa/run_schlogl_opencl.py
```diff
@@ -0,0 +1,61 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+from pysb.examples.schlogl import model
+from pysb.integrate import odesolve
+from pysb.simulator.opencl_ssa import OpenCLSimulator
+
+tspan = np.linspace(0, 100, 101)
```
alubbock: Should go in `main`

pysb/examples/ras_camp_pka.py
```diff
@@ -0,0 +1,172 @@
+from pysb import *
```
alubbock: Please add a description and reference for the model

pysb/examples/schlogl.py
```diff
@@ -0,0 +1,20 @@
+from pysb import *
```
alubbock: Please add a description and reference for the model

pysb/pathfinder.py
```diff
@@ -77,7 +89,21 @@
...
+        'name': 'NVIDIA CUDA compiler',
+        'executable': {
+            'posix': 'nvcc',
+            'nt': 'nvcc.exe'
+        },
+        'env_var': 'CUDAPATH',
+        'search_paths': {
+            'posix': ('/usr/local/cuda/bin', '/usr/bin', '/opt/cuda/bin'),
+            'nt': (
+            r'C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0\bin',)
```
alubbock: Probably better to let the user use `CUDAPATH` than to hard code a location here. Particularly now since the user's `PATH` is searched.

pysb/simulator/bng.py
```diff
@@ -1,12 +1,14 @@
-from pysb.simulator.base import Simulator, SimulationResult, SimulatorException
-from pysb.bng import BngFileInterface, load_equations, generate_hybrid_model
-import numpy as np
+import collections
```
alubbock: In a few files you've got rearranged imports. Unless there's a reason, ideally could you revert this?

pysb/simulator/cuda_ssa.py
```diff
@@ -0,0 +1,254 @@
...
+import os
+import time
+from pysb.logging import setup_logger
+import logging
+from pysb.pathfinder import get_path
+from pysb.simulator.base import SimulationResult, SimulatorException
+from pysb.simulator.ssa_base import SSABase
+
+
+class CUDASimulator(SSABase):
```
alubbock: `CudaSSASimulator` might be clearer/more specific as a name

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,236 @@
...
+    cl = None
+    device_type = None
+    ocl_array = None
+
+from pysb.bng import generate_equations
+from pysb.simulator.base import SimulationResult, SimulatorException
+from pysb.simulator.cuda_ssa import SSABase
+
+
+class OpenCLSimulator(SSABase):
```
alubbock: `OpenCLSSASimulator` might be clearer/more specific as a name

pysb/simulator/cuda_ssa.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    CUDA simulator for NVIDIA gpus
+
+    Requires pycuda, cuda, and cuda compatible NVIDIA gpus.
```
alubbock: Repeated `cuda` here. Maybe provide a link to pycuda?

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    OpenCL simulator
+
+    Requires opencl and pyopencl.
```
alubbock: Maybe provide a link to pyOpenCL?

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,236 @@
...
+        Time values passed to the constructor.
+    """
+    _supports = {'multi_initials': True, 'multi_param_values': True}
+
+    def __init__(self, model, verbose=False, tspan=None, device='gpu',
+                 multi_gpu=False,
+                 **kwargs):
+
+        if cl is None:
+            raise SimulatorException('pyopencl library required for {}'
```
alubbock: I think this should raise an `ImportError`

pysb/simulator/cuda_ssa.py
```diff
@@ -0,0 +1,254 @@
...
+    model : pysb.Model
+        Model passed to the constructor.
+    tspan : vector-like
+        Time values passed to the constructor.
+    """
+    _supports = {'multi_initials': True, 'multi_param_values': True}
+
+    def __init__(self, model, verbose=False, tspan=None, **kwargs):
+        if pycuda is None:
+            raise SimulatorException('pycuda library required for {}'
```
alubbock: Maybe an`ImportError` would be better here?

pysb/simulator/ssa_base.py
```diff
@@ -0,0 +1,159 @@
...
+        # private attribute
+        self._parameter_number = len(self._model.parameters)
+        self._n_species = len(self._model.species)
+        self._n_reactions = len(self._model.reactions)
+        self._step_0 = True
+        self.num_sim = None
+        if verbose:
+            setup_logger(logging.INFO)
+
+    def _get_template_args(self):
```
alubbock: There's a lot of regex magic in this function. Please could you add a comment or two to describe what's going on?

pysb/simulator/ssa_base.py
```diff
@@ -0,0 +1,159 @@
...
+            initials = np.repeat(self.initials, num_sim, axis=0)
+        elif len(initials.shape) == 1:
+            initials = np.repeat([initials], num_sim, axis=0)
+        self.num_sim = num_sim
+        super(SSABase, self).run(tspan=tspan, initials=initials,
+                                 param_values=param_values,
+                                 _run_kwargs=locals())
+
+    @staticmethod
+    def _get_stoch(model):
```
alubbock: Small point, but should this be `_get_stoich`?

pysb/simulator/templates/opencl.cl
```diff
@@ -0,0 +1,149 @@
...
+#define NPARAM {n_params}
+#define NREACT {n_reactions}
+#define NREACT_MIN_ONE NREACT-1
+
+typedef double output_t;
+typedef double4 output_vec_t;
+typedef philox2x32_ctr_t ctr_t;
+typedef philox2x32_key_t key_t;
+
+#define GET_RANDOM_NUM(gen) ( ((double) 2.3283064365386963e-10) * convert_double4(gen)+ ((double) 5.421010862427522e-20) * convert_double4(gen))
```
alubbock: Could you explain the meaning of these [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))? I can mostly see what this code is doing, but a little explanation about the above and the `gen_bits` function would be helpful.

pysb/simulator/templates/opencl.cl
```diff
@@ -0,0 +1,149 @@
...
+            output_vec_t ran = GET_RANDOM_NUM(gen_bits(&k, &c));
+            double r1 = ran.s0;
+	        double r2 = ran.s1;
+
+
+            double tau = -log(r1)/a0;  // find time of next reaction
+            t += tau;  // update time
+
+            long k = sample(A, a0*r2);  // find next reaction
+//            if (tid==1){{ printf("reaction %i\ta0=%f\n", k, a0); }}
```
alubbock: Would a `debug` flag be good for this stuff, rather than commenting it out? Maybe linked to PySB's `EXTENDED_DEBUG`?

pysb/simulator/ssa_base.py
JamesPino: Good catch!

pysb/simulator/cuda_ssa.py
JamesPino: Fixed

pysb/examples/schlogl.py
JamesPino: Done

pysb/examples/schlogl.py
```diff
@@ -0,0 +1,24 @@
+""" Schlogl model
```
jmuhlich: You can put the ö back if you put `# -*- coding: utf-8 -*-` at the top of this file. Or, if you want to leave it as pure ASCII, the proper alternative written form is "Schloegl".

JamesPino: Thanks! I knew I was missing something trivial.

JamesPino:
Has anyone had time to review the changes after the comments? We are near submission for a quick app note and wanted to make sure it is good to go. Thanks!

alubbock:
@JamesPino There are a few remaining comments on the code review. Please could you either address those or leave a response? Also, it'd be helpful if could you merge against master. Thanks!

pysb/simulator/ssa_base.py
JamesPino: Added

pysb/simulator/templates/opencl.cl
JamesPino: Fixed magic numbers and added citations for RNG

pysb/simulator/templates/opencl.cl
JamesPino: Added compile time options to define VERBOSE or VERBOSE_MAX based on verbose flag

pysb/simulator/ssa_base.py
JamesPino: Added comments.

benchmarks/simulator.py
```diff
@@ -45,3 +46,26 @@ def time_scipy_lsoda_no_compiler_directives(self):
 
     def time_cupsoda(self):
         self.sim_cupsoda.run()
+
+
+class SSASchlogl(object):
```
alubbock: Should be `SSASchloegl` (with an 'e') to match the others

JamesPino: Yep! Fixed.

pysb/pathfinder.py
JamesPino: Should I remove the search paths then?

pysb/simulator/bng.py
JamesPino: Done. I found where pycharm was doing it on commits.

pysb/simulator/cuda_ssa.py
JamesPino: I agree. Changed

pysb/simulator/cuda_ssa.py
JamesPino: Done

pysb/simulator/opencl_ssa.py
JamesPino: Done.

pysb/simulator/opencl_ssa.py
JamesPino: Done

pysb/examples/gpu_ssa/run_earm_cuda.py
```diff
@@ -0,0 +1,41 @@
...
+    plt.plot(tspan, x.min(axis=1), 'k--', lw=2)
+
+    sol = ScipyOdeSimulator(model, tspan)
+    traj = sol.run()
+
+    plt.plot(tspan, np.array(traj.observables)[name], label='ode', color='red')
+
+    plt.legend(loc=0)
+    plt.tight_layout()
+    plt.savefig('example_ssa_earm_cuda.png', dpi=200, bbox_inches='tight')
```
alubbock: Should probably `print` something like "Example image saved to _whatever_.png", so the user knows what's happening

pysb/examples/gpu_ssa/run_earm_opencl.py
```diff
@@ -0,0 +1,39 @@
...
+    plt.plot(tspan, x.min(axis=1), 'k--', lw=2)
+
+    sol = ScipyOdeSimulator(model, tspan)
+    traj = sol.run()
+
+    plt.plot(tspan, np.array(traj.observables)[name], label='ode', color='red')
+
+    plt.legend(loc=0)
+    plt.tight_layout()
+    plt.show()
```
alubbock: Some examples show a plot, some save it to a file, and some do both. Should probably be consistent (maybe do both?)

pysb/examples/gpu_ssa/run_schloegl_cuda.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # create histogram
+    plt.subplot(122)
+    plt.ylim(0, 800)
+    weights = np.ones_like(x[-1, :]) / float(len(x[-1, :]))
+    plt.hist(x[-1, :], 25, orientation='horizontal', weights=weights)
+    plt.yticks([])
+    plt.tight_layout()
+    out_name = '{}.png'.format(savename)
+    plt.savefig(out_name)
```
alubbock: Confirm where the figure was saved with a `print` call

pysb/examples/gpu_ssa/run_schloegl_cuda.py
```diff
@@ -0,0 +1,93 @@
...
+    except ImportError:
+        raise ImportError("Please install imageio to create a gif")
+
+    kargs = {'duration': 8 / len(images)}
+
+    imageio.mimsave(
+        'schloegl.gif',
+        [imageio.imread(i) for i in images],
+        **kargs
+    )
```
alubbock: Confirm where the figure is saved with a `print` call

pysb/examples/gpu_ssa/run_schloegl_opencl.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # create histogram
+    plt.subplot(122)
+    plt.ylim(0, 800)
+    weights = np.ones_like(x[-1, :]) / float(len(x[-1, :]))
+    plt.hist(x[-1, :], 25, orientation='horizontal', weights=weights)
+    plt.yticks([])
+    plt.tight_layout()
+    out_name = '{}.png'.format(savename)
+    plt.savefig(out_name)
```
alubbock: Confirm where the figure is saved with a `print` call

pysb/examples/gpu_ssa/run_schloegl_opencl.py
```diff
@@ -0,0 +1,93 @@
...
+    except ImportError:
+        raise ImportError("Please install imageio to create a gif")
+
+    kargs = {'duration': 8 / len(images)}
+
+    imageio.mimsave(
+        'schloegl.gif',
+        [imageio.imread(i) for i in images],
+        **kargs
+    )
```
alubbock: Confirm where the figure is saved with a `print` call

JamesPino: Added

pysb/examples/gpu_ssa/run_schloegl_opencl.py
JamesPino: Added

pysb/examples/gpu_ssa/run_schloegl_cuda.py
JamesPino: Added

pysb/examples/gpu_ssa/run_schloegl_cuda.py
JamesPino: Added

pysb/examples/gpu_ssa/run_earm_opencl.py
JamesPino: Done

pysb/examples/gpu_ssa/run_earm_cuda.py
JamesPino: Done!

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,357 @@
...
+    Parameters
+    ----------
+    model : pysb.Model
+        Model to simulate.
+    tspan : vector-like, optional
+        Time values over which to simulate. The first and last values define
+        the time range. Returned trajectories are sampled at every value unless
+        the simulation is interrupted for some reason, e.g., due to
+        satisfaction
+        of a logical stopping criterion (see 'tout' below).
```
FFroehlich, line 52: no tout below, also don't see anything about a logical stopping - is this referring to an earlier version of the code?

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,357 @@
...
+        t_out = np.array(tspan, dtype=self._dtype)
+        # compile kernel and send parameters to GPU
+        if self._step_0:
+            self._setup()
+
+        self._logger.info("Creating content on device")
+        n_sim_per_device = self.num_sim//self._n_devices
+
+        if self.num_sim < self._local_work_size[0]:
+            local_work_size = (1, 1)
```
FFroehlich, line 219: why not use `self.num_sim` workers instead of just one?

pysb/simulator/ssa_base.py
```diff
@@ -0,0 +1,151 @@
+from __future__ import print_function
```
FFroehlich, line 1: I think we dropped python 2 support a while ago

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,357 @@
+from __future__ import print_function
```
FFroehlich, line 1: I think we dropped python 2 support a while ago

pysb/simulator/cuda_ssa.py
```diff
@@ -0,0 +1,297 @@
+from __future__ import print_function
```
FFroehlich, line 1: I think we dropped python 2 support a while ago

pysb/simulator/cuda_ssa.py
```diff
@@ -0,0 +1,297 @@
...
+
+    Parameters
+    ----------
+    model : pysb.Model
+        Model to simulate.
+    tspan : vector-like, optional
+        Time values over which to simulate. The first and last values define
+        the time range. Returned trajectories are sampled at every value unless
+        the simulation is interrupted for some reason, e.g., due to
+        satisfaction
```
FFroehlich, line 45: no tout below, also don't see anything about a logical stopping - is this referring to an earlier version of the code?

FFroehlich (dismissed):

pysb/simulator/opencl_ssa.py
jmuhlich: This is OK, it's a verbatim copy of the base class docstring. tout is part of the SimulationResult class. I agree this is a bit obscure, but it's not something that needs to be addressed in this PR.

pysb/simulator/cuda_ssa.py
jmuhlich: This is OK - see other instance of this in OpenCLSSASimulator.

pysb/simulator/opencl_ssa.py
jmuhlich: The rationale behind all these calculations around number of simulations, work size, blocks, threads, etc. is not obvious. A comment explaining _why_ (not what) would help a lot here. Then we can evaluate whether the code actually implements that design accurately.

pysb/simulator/ssa_base.py
```diff
@@ -0,0 +1,151 @@
...
+        elif isinstance(initials, np.ndarray) and len(initials.shape) == 1:
+            # parameters taken care of on SimulatorBase side
+            initials = np.repeat([initials], num_sim, axis=0)
+        self.num_sim = num_sim
+        super(SSABase, self).run(tspan=tspan, initials=initials,
+                                 param_values=param_values,
+                                 _run_kwargs=locals())
+
+    @staticmethod
+    def _get_stoich(model):
```
jmuhlich, line 131: Could you use `model.stoichiometry_matrix` instead? It's very new and it definitely wasn't available when you started this. I already noticed the code here doesn't handle fixed-concentration species so that's one feature you'd get for free by adopting it.

https://github.com/pysb/pysb/blob/master/pysb/core.py#L2109

doc/conf.py
```diff
@@ -51,17 +51,19 @@
...
+                'sympy.printing', 'sympy.printing.mathml',
+                'sympy.printing.lambdarepr', 'numpy',
+                'scipy', 'scipy.integrate', 'scipy.constants', 'scipy.sparse',
+                'h5py', 'theano', 'theano.tensor', 'dateutil',
+                'dateutil.parser', 'networkx', 'networkx.algorithms',
+                'networkx.algorithms.isomorphism',
+                'networkx.algorithms.isomorphism.vf2userfunc',
+                'nose', 'nose.tools', 'matplotlib', 'matplotlib.pyplot',
+                'concurrent', 'concurrent.futures', 'pycuda.driver', 'pycuda',
+                'pyopencl']
```
jmuhlich, line 65: I know the formatting on this is not ideal, but please avoid reformatting it dramatically as this has introduced a merge conflict. Adding back in the extra column of spaces removes the conflict so please do that before merging master into this branch again (required for final approval). Also it would be nice to avoid rewrapping the lines you didn't edit to eliminate noise from the diff. Reformatting large blocks of code should be done in its own PR.

jmuhlich (changes requested):
Looks good overall, see individual comments. One overarching thing to fix is to please ensure all modified files have a line ending character on the last line.

After addressing all comments you will need to resolve the merge conflict (see comment on doc/conf.py for a quick fix for this), merge from master one last  time, and wait for the CI run to pass before we merge this.

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,357 @@
...
+            try:
+                traj = [
+                    r.result().reshape(
+                        (total_threads, len(t_out), self._n_species)
+                    )
+                    for r in results
+                ]
+            finally:
+                for r in results:
+                    r.cancel()
```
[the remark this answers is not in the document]
JamesPino: I think I had some issues with multi-GPUs (using a regular pool, not futures) and lagging out and was trying to come up with a solution. I don't know why I tried this, nor how I fixed it without, but not needed for sure.  Removed

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,357 @@
...
+        traj = traj[:self.num_sim]
+        self._time = time.time() - timer_start
+        self._logger.info("{} simulations "
+                          "in {:.4f}s".format(self.num_sim, self._time))
+
+        tout = np.array([tspan] * self.num_sim)
+        return SimulationResult(self, tout, traj)
+
+    @staticmethod
+    def get_blocks(n_simulations, threads_per_block):
```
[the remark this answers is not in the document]
JamesPino: Good catch. Refactored.

pysb/simulator/opencl_ssa.py
```diff
@@ -0,0 +1,357 @@
...
+        # transfer the array of time points to the device
+        time_points_gpu = ocl_array.to_device(
+            queue,
+            np.array(t_out, dtype=ocl_instance._dtype)
+        )
+        # total_threads = int(self.num_sim)
+        # transfer the array of seeds to the device
+        random_seeds_gpu = ocl_array.to_device(
+            queue,
+            np.array(random.sample(range(2 ** 32 - 1), total_threads),
```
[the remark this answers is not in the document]
JamesPino: Good point. CUDA could use clock64() to set a seed, so I started there and kind of left it. Then I made the openCL version, which required a seed (no clock available in kernel). I didn't think twice, but completely agree. I made it where either one can now be passed a seed, which initiates the rng state. This works the same for both simulators, however if multiple gpus is used with openCL, then the rng state is offset by the device number. 
Its based off https://albertcthomas.github.io/good-practices-random-number-generators/

pysb/simulator/ssa_base.py
JamesPino: Nice! Updated

doc/conf.py
JamesPino: Reverted back.

benchmarks/simulator.py
```diff
@@ -45,3 +46,26 @@ def time_scipy_lsoda_no_compiler_directives(self):
...
+
+class SSASchloegl(object):
+    def setup(self):
+        self.nsims = 100
+        self.timer = timeit.default_timer
+        self.model = schloegl.model
+        self.tspan = np.linspace(0, 100, 101)
+        self.stochkit_sim = StochKitSimulator(self.model, tspan=self.tspan)
+        self.cuda_ssa_sim = CudaSSASimulator(self.model, tspan=self.tspan)
+        self.opencl_ssa_sim = OpenCLSSASimulator(self.model, tspan=self.tspan)
```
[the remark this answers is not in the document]
JamesPino: Updated. Is expensive since it has to compile everything the first time.

pysb/simulator/opencl_ssa.py
JamesPino: Added. Despite my best attempts, there is not an "exact" optimal number for work size, blocks, and number of threads. In an ideal world, when the GPUs are being used properly with no branching, there is. But in our case, there are ball park best, but the branching makes it trivial optimizations overall. I added the important parts, which basically try to make sure chunks of simulations go through as a warp size of 32 for NVIDIA, or wavefront size of 64 for AMD, but what really matters is that the work size or threads per block is a multiple of these. 
For small number of simulations (small than 32 or 64), it really doesn't matter what value you set, so I set to 1. But to simplify, I just reverted to be local_work_size. Since it generates up to a block, it pads the rest with 0s for either case.

JamesPino:
Seems to be failing based on kappa simulator. Don't think I touched it? Will dive deeper later.

benchmarks/simulator.py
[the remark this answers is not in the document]
jmuhlich: Check the ASV docs on `setup_cache` -- it needs to return something, and the `time_*` methods also need to be refactored to accept that same thing. I think this benchmark will error as written now, or it might just give weird/wrong results. You can go back to using plain `setup()` and it will be slow but correct, and we can resolve this some other time.

jmuhlich (changes requested):
One minor thing to fix up with the benchmark suite, but looks good other than that.

JamesPino:
@jmuhlich I keep trying to figure out where github actions is failing, but am unable to figure it out. It looks like it is from simulationResult when setting tout. Locally it works. On the Lolab-msm gh-action it worked for windows, but failed for ubuntu-latest. https://github.com/LoLab-MSM/pysb/actions/runs/4547289909/jobs/8016989175

image

Any thoughts? I am thinking the error must be from something else, which triggers a cancel workflow, and those are the errors I am getting.

alubbock:
@JamesPino 
> I keep trying to figure out where github actions is failing, but am unable to figure it out.

A change in numpy v1.24 (specifically numpy/numpy#22004) broke the tests due to a change in how ragged arrays are handled. It's a simple fix, which I've put in a new PR, #574.

jmuhlich (approved):

alubbock (approved):

FFroehlich (approved):

PR #447, opened 2019-04-26
KappaSimulator

alubbock:
A simulation engine for Kappa built on the PySB Simulator class.
Supports multiple realizations, and multiple initial condition
and parameter sets.

This PR is rebased off #445 which is required for tests to pass.


pysb/simulator/kappa.py
```diff
@@ -0,0 +1,178 @@
...
+                        '-i', kappa_filename, '-o', out_filename]
+
+                    # Run KaSim
+                    self._logger.debug('Running: {}'.format(' '.join(args)))
+                    p = subprocess.Popen(args,
+                                         stdout=subprocess.PIPE,
+                                         stderr=subprocess.PIPE,
+                                         cwd=base_directory)
+
+                    for line in iter(p.stdout.readline, b''):
```
jmuhlich: Does this do something different here than the shorter `for line in p.stdout` ?

pysb/tests/test_simulator_kappa.py
```diff
@@ -0,0 +1,66 @@
...
+
+def _compare_kappa_sims(run_simulation_output, kappasimulator_output):
+    sim1 = run_simulation_output.view('<f8')[:, 1:]
+    sim2 = kappasimulator_output.observables.view('f8').reshape(sim1.shape)
+
+    assert np.allclose(sim1, sim2, atol=1e-16, rtol=1e-16)
+
+
+def test_kappa_sim_michment():
+    orig_sim = run_simulation(michment.model, time=100, points=100, seed=123)
```
jmuhlich: I figured we would deprecate `pysb.kappa.run_simulation` as part of this change. If we do that, it probably doesn't make sense to test against it. Thoughts?

pysb/simulator/kappa.py
alubbock: No, `for line in p.stdout` seems to work too. I'll switch to that.

pysb/tests/test_simulator_kappa.py
alubbock: I've now added perturbation support to `KappaSimulator` and deprecated `pysb.kappa.run_simulation`. I don't think there's any need to add flux map support, since users can use the dedicated function for that.

I'd suggest leaving the unit tests in place for deprecated code until it's actually removed.

alubbock: I see; apologies for the misunderstanding. I'll work on a new PR to remove the tests' dependence on deprecated code before the PySB 2 update.

PR #451, opened 2019-07-10
include macros functions in rule._function

ortega2247:
Changed the code to include macros functions in the _function attribute of the Component class. This could be helpful as a way to annotate rules with the macros that generate them.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -191,7 +191,7 @@ class Component(object):
             mod_name = frame.f_globals.get('__name__', '__unnamed__')
             if mod_name in ['IPython.core.interactiveshell', '__main__']:
                 break
-            if mod_name not in ['pysb.core'] and not \
+            if mod_name != 'pysb.core' and not \
                     mod_name.startswith('importlib.'):
                 self._modules.append(mod_name)
                 if self._function is None:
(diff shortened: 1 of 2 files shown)


pysb/core.py
```diff
@@ -191,11 +191,14 @@ def __init__(self, name, _export=True):
             mod_name = frame.f_globals.get('__name__', '__unnamed__')
             if mod_name in ['IPython.core.interactiveshell', '__main__']:
                 break
-            if mod_name not in ['pysb.core', 'pysb.macros'] and not \
+            if mod_name not in ['pysb.core'] and not \
```
alubbock: Can use `!=` here

ortega2247: Done

jmuhlich:
This makes me think this feature should be under more user control, as some might want macros included, some might not, and some might indeed want the whole stack. We shouldn't try to make the existing code any hairier than it already is, but in a future version maybe we can work out a way to reengineer this feature "orthogonally" to the Component class itself.

alubbock:
Agreed @jmuhlich, probably best to capture the whole function stack in a future version for maximum flexibility. Not sure how it could be done outside of the class itself - maybe some kind of hook mechanism?

For now, this LGTM. Thanks @ortega2247.

PR #452, opened 2019-07-11
Fix ModelNotDefinedError

alubbock:
This error should be raised when self exporter is used and a model
component is declared before Model(). It's broken on
Python 3.7 without this PR:

    TypeError: descriptor '__init__' requires a 'ValueError' object but received a 'ModelNotDefinedError'


jmuhlich:
By the way it breaks on 2.7 as well.

PR #456, opened 2019-08-13
Improve BioNetGen function expression parsing

jmuhlich:
BNGL models (and also SBML due to the conversion path) with piecewise
expressions and logical/equality expressions will now parse into
sympy expressions correctly. Piecewise expressions ('if' in BNGL)
in particular would trigger an exception in the sympy parser.

### DIFF
--- a/pysb/importers/bngl.py
+++ b/pysb/importers/bngl.py
@@ -4,7 +4,8 @@ from pysb.builder import Builder
 from pysb.bng import BngFileInterface
 import xml.etree.ElementTree
 import re
-from sympy.parsing.sympy_parser import parse_expr
+import sympy
+import sympy.parsing.sympy_parser as sympy_parser
 import warnings
 import pysb.logging
 import collections
@@ -65,7 +66,7 @@ class BnglBuilder(Builder):
 
         # Quick security check on the expression
         if re.match(r'^[\w\s()/+\-._*]*$', expression):
-            return parse_expr(expression, self._model_env)
+            return parse_bngl_expr(expression, self._model_env)
         else:
             self._warn_or_except('Security check on expression "%s" failed' %
                                  expression)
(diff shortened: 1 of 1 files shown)


jmuhlich:
Suggestions on a testing strategy for this code are welcome. Should we cook up a torture test SBML/BNGL file or just test the new function I added that now contains all the conversion logic?

alubbock:
I'd suggest using a token conversion function rather than string replacement, wherever possible.

Here's my attempt, by modifying `parse_bngl_expr` and adding `_convert_tokens`:

```python
def _convert_tokens(tokens, local_dict, global_dict):
    from tokenize import NAME, OP
    for pos, tok in enumerate(tokens):
        if tok == (NAME, 'if'):
            tokens[pos] = (NAME, '__bngl_if')
            local_dict['__bngl_if'] = sympy.Function('bngl_if')
        elif tok == (OP, '^'):
            tokens[pos] = (OP, '**')
        elif tok == (NAME, 'and'):
            tokens[pos] = (OP, '&')
        elif tok == (NAME, 'or'):
            tokens[pos] = (OP, '|')
    return tokens


def parse_bngl_expr(text, *args, **kwargs):
    """Convert a BNGL math expression string to a sympy Expr."""

    # Translate a few operators with simple text replacement.
    text = text.replace('()', '')
    text = text.replace('==', '=')
    # Use sympy to parse the text into an Expr.
    trans = (
        sympy_parser.standard_transformations
        + (sympy_parser.convert_equals_signs, _convert_tokens)
    )
    expr = sympy_parser.parse_expr(text, *args, transformations=trans, **kwargs)
    # Transforming 'if' to Piecewise requires subexpression rearrangement, so we
    # use sympy's replace functionality rather than attempt it using text
    # replacements above.
    expr = expr.replace(
        sympy.Function('bngl_if'),
        lambda cond, t, f: sympy.Piecewise((t, cond), (f, True))
    )
    # Check for unsupported constructs.
    if expr.has('time'):
        raise ValueError(
            "Expressions referencing simulation time are not supported"
        )
    return expr
```

alubbock:
For testing, here are a few unit tests as a start:

```python
def test_parse_bngl_expression_if():
    x, y = sympy.symbols('x y')
    assert parse_bngl_expr('if(x>y, 1, 3)') == \
        sympy.Piecewise((1, x > y), (3, True))


def test_parse_bngl_expression_exponentiate():
    x, y = sympy.symbols('x y')
    assert parse_bngl_expr('x ^ y') == x ** y


def test_parse_bngl_expression_and_or_equals():
    x, y = sympy.symbols('x y')
    assert parse_bngl_expr('x and y') == sympy.And(x, y)
    assert parse_bngl_expr('x or y') == sympy.Or(x, y)
    assert parse_bngl_expr('x == y') == sympy.Eq(x, y)
```

alubbock:
There's also an issue where BioNetGen's parser is more weakly typed than sympy: Booleans get automatically converted to integers in BNG, but not in sympy, so an expression like `(10 > 5) * 4` fails on import, but would get evaluated to `4` by BNG. I've attempted a fix for that in the `_fix_booleans` function in #412, since this issue crops up a lot with local functions.

alubbock:
Included and extended in #412

PR #457, opened 2019-08-14
Evaluate expression

ortega2247:
Use parameter and expression values (when the expression is constant) to calculate the value of a defined expression.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1542,7 +1542,7 @@ class Expression(Component, sympy.Symbol):
                 subs[a] = a.value
             elif isinstance(a, Expression) and a.is_constant_expression():
                 subs[a] = a.get_value()
-        return self.expr.evalf(subs=subs)
+        return self.expr.xreplace(subs)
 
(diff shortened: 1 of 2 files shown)


pysb/core.py
```diff
@@ -1535,7 +1535,14 @@ def is_constant_expression(self):
...
     def get_value(self):
-        return self.expr.evalf()
+        # Use parameter and expression values for evaluation
+        subs = {}
+        for a in self.expr.atoms():
+            if isinstance(a, Parameter):
+                subs[a] = a.value
+            elif isinstance(a, Expression) and a.is_constant_expression():
+                subs[a] = a.get_value()
+        return self.expr.evalf(subs=subs)
```
jmuhlich: Suggest `xreplace(subs)` instead of `evalf(subs=subs)` here. It implements just the replacement functionality we want, and it's much faster than `evalf`.

ortega2247: Used xreplace as suggested. Thanks!

PR #458, opened 2019-08-14
Add dashed edge when reaction rate is an expression with observables

ortega2247:


ortega2247:
Added  the keyword argument include_rate_species and the command line option as suggested. One think that I noticed is that for models in which a species is in the reaction reactants and in the reaction rate there would be only one edge between that species node and the reaction node. We could allow multiple edges between nodes, when include_rate_species=True, if we make the graph a multigraph. Thoughts??

alubbock:
I think a multigraph is OK here. It's a more complete visualisation, and can always be simplified to a graph if needed.

The alternative would be to represent combined reaction/rate relationships another way, for example using a different type of edge (e.g. dash-dot).

PR #464, opened 2019-10-07
Local function export fixes, example model

alubbock:
* New `localfunc.py` example model, based on a BioNetGen example model
* Fix export of models with local functions to SBML and StochKit
* Show meaningful error messages when attempting to export models
with local functions to cupSODA or Kappa
* Fix local function parameter reads from BioNetGen when parameter
is actually a constant expression, e.g. `2^2`

### DIFF
--- a/pysb/simulator/cupsoda.py
+++ b/pysb/simulator/cupsoda.py
@@ -15,7 +15,7 @@ import shutil
 from pysb.pathfinder import get_path
 import sympy
 import collections
-
+import itertools
 try:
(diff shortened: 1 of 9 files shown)


pysb/simulator/cupsoda.py
```diff
@@ -15,7 +15,7 @@
 from pysb.pathfinder import get_path
 import sympy
 import collections
-
+import itertools
```
jmuhlich: This seems spurious.

Issue #467, opened 2019-10-17
New tags attribute causes error when unpickling older models


bgyori:
The PR https://github.com/pysb/pysb/pull/464/ adds a new `tags` attribute to the Model class. For any model that was pickled before this PR was merged, we get an error in the setstate code:

```python
    def __setstate__(self, state):
        # restore the 'model' weakrefs on all components
        self.__dict__.update(state)
        for c in self.all_components():
            c.model = weakref.ref(self)
```
like 
```
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pysb/core.py", line 1845, in __setstate__
    for c in self.all_components():
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pysb/core.py", line 1914, in all_components
    for cset in self.all_component_sets():
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pysb/core.py", line 1908, in all_component_sets
    sets = [getattr(self, name) for name in set_names]
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pysb/core.py", line 1908, in <listcomp>
    sets = [getattr(self, name) for name in set_names]
AttributeError: 'Model' object has no attribute 'tags'
```
Should anything in PySB be done to mitigate this, or is this purely a downstream concern?

alubbock:
This should get fixed in PySB. Apologies for the oversight - fortunately it's a one-line fix. See #468.

jmuhlich:
There are going to be lots of other problems using a Model unpickled from an older PySB version. For example @bgyori's unpickled models will crash the simulators since _derived_parameters and _derived_expressions won't exist. The problem isn't even limited to this PR or even to the Model class -- any class that changes its structure will be a problem. See here:

https://wiki.python.org/moin/UsingPickle#Discussion

> Pickles can cause problems if you save a pickle, then update your code and read the pickle in. Attribute added to your \_\_init\_\_ may not be present in the unpickled object ... For this reason, you should be wary of using pickles for long-term storage where the underlying code is not highly stable.

I think handling this sort of thing does belong in the end-user's code. Maybe you could write a pickle upgrade script that monkey-patches Model.__setstate__ to add all the new attributes?

bgyori:
[...]

PR #468, opened 2019-10-17
Fix unpickling of models saved in PySB <1.10

alubbock:
`tags` attribute is not set on models saved in PySB <1.10,
which breaks unpickling. We fix this by setting the attribute
if not already present.

Fixes: #467

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1842,6 +1842,8 @@ class Model(object):
     def __setstate__(self, state):
         # restore the 'model' weakrefs on all components
         self.__dict__.update(state)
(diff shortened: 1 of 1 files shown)


bgyori:
Thanks! If you don't mind making a new minor release on PyPI to propagate this once it's merged, that would be useful.

jmuhlich:
Isn't this a bigger problem than just this one attribute? I don't think pickle/unpickle is expected to work across different versions of a class. Will take discussion back to #467.

alubbock:
Closed in favour of #469

PR #469, opened 2019-10-19
JSON model import and export

alubbock:
Add JSON model import and export capabilities. JSON
is a text-serializable format, which is good for storing
models within other objects, or for purposes like HTTP
APIs.

SimulationResult.save() now saves models in JSON
rather than pickle.

See discussion on #467 which prompted the need for
a durable and serializable model storage format.

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1271,8 +1271,16 @@ class Parameter(Component, sympy.Symbol):
         return (self.name, self.value, False)
 
     def __init__(self, name, value=0.0, _export=True):
-        self.value = float(value)
+        self.value = value
         Component.__init__(self, name, _export)
+
+    @property
+    def value(self):
+        return self._value
+
+    @value.setter
+    def value(self, new_value):
+        self._value = float(new_value)
     
     def get_value(self):
         return self.value
--- a/pysb/examples/bngwiki_egfr_simple.py
+++ b/pysb/examples/bngwiki_egfr_simple.py
@@ -39,7 +39,7 @@ Parameter('deg', 0.01)            # degradation of receptor dimers
 
 
 Monomer('EGF', ['R'])
-Monomer('EGFR', ['L','CR1','Y1068'], {'Y1068':('U','P')})
+Monomer('EGFR', ['L','CR1','Y1068'], {'Y1068': ['U', 'P']})
 Monomer('Grb2', ['SH2','SH3'])
 Monomer('Sos1', ['PxxP'])
 
--- a/pysb/export/json.py
+++ b/pysb/export/json.py
@@ -39,7 +39,7 @@ class JsonExporter(Exporter):
 
 
 class PySBJSONEncoder(json.JSONEncoder):
-    FORMAT = 1
+    PROTOCOL = 1
 
     @classmethod
     def encode_keyword(cls, keyword):
@@ -164,7 +164,7 @@ class PySBJSONEncoder(json.JSONEncoder):
 
     @classmethod
     def encode_model(cls, model):
-        d = dict(format=cls.FORMAT, name=model.name)
+        d = dict(protocol=cls.PROTOCOL, name=model.name)
 
         encoders = {
             'monomers': cls.encode_monomer,
--- a/pysb/export/json.py
+++ b/pysb/export/json.py
@@ -7,8 +7,9 @@ for :py:mod:`pysb.export`.
 
 from __future__ import absolute_import
 from pysb.export import Exporter
+from pysb.bng import generate_equations
 import json
-from pysb.core import Model, MultiState, KeywordMeta
+from pysb.core import Model, MultiState, KeywordMeta, Parameter, Expression
 
 
 class JsonExporter(Exporter):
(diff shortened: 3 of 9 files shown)


bgyori:
This looks promising! One of the issues though, is that often, one of the main rationales behind pickling a PySB model is to save the result of having run bng.generate_equations on the model (which for certain models can be quite slow). The model can then be quickly reloaded and simulated without having to regenerate the reaction network. This is not immediately possible with the JSON approach, though perhaps could be optionally handled by encoding the equations in JSON as well.

alubbock:
@bgyori No problem, I've added an option for this: `JsonExporter(model).export(include_netgen=True)`

pysb/tests/test_exporters.py
```diff
@@ -101,6 +106,13 @@ def check_convert(model, format):
                     print(pd.DataFrame(dict(rr=rr_obs, pysb=py_obs)))
                     raise ValueError('Model {}, observable__o{} "{}" trajectories do not match:'.format(
                         model.name, obs_idx, model.observables[obs_idx].name))
+        elif format == 'json':
+            # Round-trip the model by re-importing the JSON
+            m = model_from_json(exported_file)
+            # Check network generation and single-step integrator
+            if model.name not in ('pysb.examples.tutorial_b',
+                                  'pysb.examples.tutorial_c'):
+                ScipyOdeSimulator(m)
```
jmuhlich: This is a reasonable smoke test to make sure the model isn't broken, but it doesn't demonstrate we got the same model back. We ought to actually check that the model is the same, but that's probably infeasible on every example unless we leverage some other exporter (and hope we don't end up with parallel bugs in both exporters). The first thing that comes to mind is to run each model and its re-imported version through the pysb_flat exporter and compare the strings. Of course the pysb_flat exporter doesn't have a round-trip fidelity test either, does it?

Or maybe create a minimal model that exercises every feature once, and explicitly check all the components before and after re-import? That might be a big job, though.

pysb/simulator/base.py
```diff
@@ -1288,3 +1309,24 @@ def _allow_unicode_recarray():
...
     return True
+
+
+def _model_setstate_monkey_patch(self, state):
+    """Monkey patch for Model.__setstate__ for restoring from older pickles"""
+
+    # restore the 'model' weakrefs on all components
+    self.__dict__.update(state)
+    # Set "tags" attribute for older, pickled models
+    self.__dict__.setdefault('tags', ComponentSet())
```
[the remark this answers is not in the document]
alubbock: This is the only change required to unpickle models or load HDF5 files, at least as far back as PySB v1.6.0 where the HDF5 format was introduced.

For the standalone pickle cases, the following snippet works. Let me know if you think I should remove the underscore prefix and/or document this somehwere explicitly.

```python
from pysb.simulator.base import _patch_model_setstate

with open('earm_1_0_pysb1.6.pkl', 'rb') as f:
    with _patch_model_setstate():
        pickle.load(f)
```

pysb/tests/test_exporters.py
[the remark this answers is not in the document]
alubbock: The integrator RHS is run for `t=0` during initialization, at least when using Cython (line 219 of scipyode.py). But I'll change this to enforce Cython, using `compiler='cython'` kwarg.

pysb/export/json.py
```diff
@@ -0,0 +1,228 @@
...
+        -------
+        string
+            The JSON output for the model.
+        """
+        return json.dumps(self.model, cls=PySBJSONWithNetworkEncoder
+                          if include_netgen else PySBJSONEncoder)
+
+
+class PySBJSONEncoder(json.JSONEncoder):
+    FORMAT = 1
```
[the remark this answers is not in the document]
alubbock: I've switched to `protocol`, rather than `format`.

I've added some documentation to cover the basics of the format. The protocol number should get bumped when new features are added which affect model simulation, or prevent it being loaded by the decoder. I don't think it's a bad thing if this happens, so long as the decoder maintains backwards compatibility.

pysb/tests/test_exporters.py
alubbock: Indeed - I used `PySBFlatExporter` to manually check things were working as expected, but symbols within expressions don't appear in a deterministic order. I agree the best solution is probably to create a toy model that uses every feature once and check equality manually.

alubbock: I ended up using `check_model_against_component_list` on all of the example models after round-tripping to JSON, which seems to do the job.

pysb/core.py
```diff
@@ -1271,8 +1271,16 @@ def __getnewargs__(self):
         return (self.name, self.value, False)
 
     def __init__(self, name, value=0.0, _export=True):
-        self.value = float(value)
+        self.value = value
         Component.__init__(self, name, _export)
+
+    @property
+    def value(self):
```
jmuhlich: What's this new getter/setter for? I'm guessing round-trip testing exposed a code path where the normalization of `.value` to a float was getting bypassed?

pysb/examples/bngwiki_egfr_simple.py
```diff
@@ -39,7 +39,7 @@
 
 
 Monomer('EGF', ['R'])
-Monomer('EGFR', ['L','CR1','Y1068'], {'Y1068':('U','P')})
+Monomer('EGFR', ['L','CR1','Y1068'], {'Y1068': ['U', 'P']})
```
jmuhlich: Maybe Monomer should just normalize `sites` and `site_states.values()` to `list`, but that can be done outside of this PR.

pysb/tests/test_exporters.py
```diff
@@ -112,7 +114,12 @@ def check_convert(model, format):
             # Check network generation and single-step integrator
```
jmuhlich: As you pointed out in your previous comment, we actually just need to evaluate the RHS once which is different from stepping the integrator. This comment would be more accurate as something like "Check network generation and force RHS evaluation".

pysb/core.py
alubbock: Yes, in the earm_1_3.py model where `.value` is used with integers, but these values were getting converted to floats after round-tripped to JSON.

pysb/examples/bngwiki_egfr_simple.py
alubbock: Agree, we should normalize in `core.py`.

pysb/tests/test_exporters.py
alubbock: Comment fixed as requested

PR #473, opened 2019-12-10
Amici Simulator

FFroehlich:
Implements support for https://github.com/ICB-DCM/AMICI as Simulator class


FFroehlich:
Test for correctness of solutions is implemented here: https://github.com/ICB-DCM/AMICI/blob/a4e077f317d2d3eb6f93fcb1a849ddb4ffa7a175/tests/testPYSB.py#L148

FFroehlich:
@jmuhlich this is now done from my side, tests currently do not pass for python2 which is EOL.

FFroehlich:
@jmuhlich @alubbock as this PR is approaching its one year anniversary, could we decide on it's fate?

JamesPino:
Just curious so I know for later, but how come you have the pysb2amici converted in the amici repo rather than the pysb repo? Excited to do some benchmarks

FFroehlich:
> Just curious so I know for later, but how come you have the pysb2amici converted in the amici repo rather than the pysb repo? Excited to do some benchmarks

Amici doesn't have any specific file format that it can read in, so the function doesn't work like the other exporters, but rather constructs `ODEModel` and `ODEExporter` objects (see https://github.com/AMICI-dev/AMICI/blob/6aa87c34ab80c5a5e6be40b4bcccf3386e316047/python/amici/ode_export.py) and relies on some AMICI internals. Felt more natural to have this in AMICI, especially as that interface was more fluid initially, but this could of course be changed.

FFroehlich:
> Just curious so I know for later, but how come you have the pysb2amici converted in the amici repo rather than the pysb repo? Excited to do some benchmarks

We did some benchmarks against the SciPySimulator using Cython for 17 of the pysb examples:

image

PR #474, opened 2019-12-12
fix usage of super in multiple inheritance of sympy.Symbol

FFroehlich:
`super(sympy.Symbol, cls).__new__` calls `sympy.Basic.__new__`, not `sympy.Symbol.__new__`

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1265,7 +1265,7 @@ class Parameter(Component, sympy.Symbol):
     """
 
     def __new__(cls, name, value=0.0, _export=True):
-        return super(sympy.Symbol, cls).__new__(cls, name)
+        return super(Parameter, cls).__new__(cls, name)
 
     def __getnewargs__(self):
         return (self.name, self.value, False)
@@ -1545,7 +1545,7 @@ class Observable(Component, sympy.Symbol):
     """
 
     def __new__(cls, name, reaction_pattern, match='molecules', _export=True):
-        return super(sympy.Symbol, cls).__new__(cls, name)
+        return super(Observable, cls).__new__(cls, name)
 
     def __getnewargs__(self):
         return (self.name, self.reaction_pattern, self.match, False)
@@ -1613,7 +1613,7 @@ class Expression(Component, sympy.Symbol):
     """
 
     def __new__(cls, name, expr, _export=True):
-        return super(sympy.Symbol, cls).__new__(cls, name)
+        return super(Expression, cls).__new__(cls, name)
 
     def __getnewargs__(self):
         return (self.name, self.expr, False)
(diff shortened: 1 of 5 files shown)


pysb/core.py
```diff
@@ -153,6 +153,16 @@ def rename(obj, new_name):
                              "name '%s'" % obj.name)
 
 
+class Symbol(sympy.Dummy):
+    def __new__(cls, name, value=0.0, _export=True):
```
jmuhlich: Could the `value` and `_export` args be removed here?

pysb/pattern.py
```diff
@@ -928,7 +928,7 @@ class ReactionPatternMatcher(object):
     [Rxn (reversible):
         Reactants: {'__s15': mSmac(b=None), '__s45': AMito(b=None)}
         Products: {'__s47': AMito(b=1) % mSmac(b=1)}
-        Rate: __s15*__s45*kf21 - __s47*kr21
+        Rate: kf21*__s15*__s45 - kr21*__s47
```
jmuhlich: Are commutative terms automatically and consistently sorted by sympy? I'm wondering if this new term ordering is stable.

pysb/tests/test_importers.py
```diff
@@ -67,7 +67,7 @@ def bngl_import_compare_simulations(bng_file, force=False,
             renamed_species = 'Obs_{}'.format(species)
         logger.debug(yfull2[renamed_species])
         assert numpy.allclose(yfull1[species], yfull2[renamed_species],
-                              atol=precision, rtol=precision)
+                              atol=precision*10, rtol=precision*10)
```
jmuhlich: Can you explain this a bit more? Maybe changing the default `1e-12` to `1e-11` here (and on line 189) is a cleaner option. Also the values in the `REDUCED_PRECISION` dict above might need to be changed too.

pysb/pattern.py
FFroehlich: Yes, according to documentation ordering is automatic, consistent and arbitrary. See https://docs.sympy.org/1.5.1/tutorial/manipulation.html

> This because in SymPy, the arguments of the commutative operations Add and Mul are stored in an arbitrary (but consistent!) order, which is independent of the order inputted (if you’re worried about noncommutative multiplication, don’t be. In SymPy, you can create noncommutative Symbols using Symbol('A', commutative=False), and the order of multiplication for noncommutative Symbols is kept the same as the input). Furthermore, as we shall see in the next section, the printing order and the order in which things are stored internally need not be the same either.

pysb/core.py
FFroehlich: yes! removed.

pysb/tests/test_importers.py
FFroehlich: looks like only the repressilator example was failing so I added a specific case to `REDUCED_PRECISION` and reverted the above change.

PR #475, opened 2019-12-12
Symbolic assumptions

FFroehlich:
Implements the assumption that all symbolic variables in a model only take real values. Enables the user to specify integer and non-negative assumptions about parameter values and enforces them when setting respective values. Changes are relevant for any simplifications of symbolic expressions in the model.

Requires #474

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1257,6 +1257,14 @@ class Parameter(Component, sympy.Symbol):
         The numerical value of the parameter. Defaults to 0.0 if not specified.
         The provided value is converted to a float before being stored, so any
         value that cannot be coerced to a float will trigger an exception.
+    nonnegative : bool, optional
+        Sets the assumption whether this parameter is nonnegative (>=0).
+        Affects simplifications of expressions that involve this parameter.
+        By default, parameters are assumed to be non-negative.
+    integer : bool, optional
+        Sets the assumption whether this parameter takes integer values,
+        which affects simplifications of expressions that involve this
+        parameter. By default, parameters are assumed to take integer values.
 
     Attributes
     ----------
@@ -1264,15 +1272,16 @@ class Parameter(Component, sympy.Symbol):
 
     """
 
-    def __new__(cls, name, value=0.0, positive=True, integer=False,
+    def __new__(cls, name, value=0.0, nonnegative=True, integer=False,
                 _export=True):
-        return super(Parameter, cls).__new__(cls, name, real=True, positive=positive,
-                                    integer=integer)
+        return super(Parameter, cls).__new__(cls, name, real=True,
+                                             nonnegative=nonnegative,
+                                             integer=integer)
 
     def __getnewargs__(self):
         return (self.name, self.value, False)
 
-    def __init__(self, name, value=0.0, _export=True):
+    def __init__(self, name, value=0.0, _export=True, **kwargs):
         self.value = value
         Component.__init__(self, name, _export)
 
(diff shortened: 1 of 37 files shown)


pysb/core.py
```diff
@@ -154,8 +154,9 @@ def rename(obj, new_name):
 
 
 class Symbol(sympy.Dummy):
-    def __new__(cls, name):
-        return super(Symbol, cls).__new__(cls, name)
+    def __new__(cls, name, **kwargs):
+        real = kwargs.pop('real', True)
```
jmuhlich: I think adding `real=True` to the `__new__` args instead would be cleaner.

pysb/core.py
```diff
@@ -1267,20 +1268,31 @@ class Parameter(Component, Symbol):
         The numerical value of the parameter. Defaults to 0.0 if not specified.
         The provided value is converted to a float before being stored, so any
         value that cannot be coerced to a float will trigger an exception.
+    nonnegative : bool, optional
```
jmuhlich: Note that the Attributes section below currently says "Identical to Parameters", but these new `nonnegative` and `integer` parameters are not stored as attributes. Maybe that text just needs to changed to something like "Identical to Parameters, excluding \`nonnegative\` and \`integer\`", or it could say "value (see Parameters above)".

pysb/core.py
FFroehlich: done

pysb/core.py
FFroehlich: done

FFroehlich:
added consistency check to constructor and simulator.

I don't think information can be properly propagated to exporters due to inexistant support in most formats.

pysb/core.py
```diff
@@ -1302,25 +1302,31 @@ def value(self):
...
     def __repr__(self):
-        return  '%s(%s, %s)' % (self.__class__.__name__, repr(self.name), repr(self.value))
+        return '%s(%s, %s)' % (self.__class__.__name__, repr(self.name), repr(self.value))
 
     def __str__(self):
-        return  repr(self)
+        return repr(self)
+
+
+def _check_value_consistency(value, is_integer, is_nonnegative):
```
jmuhlich: I like the idea of refactoring the checking logic out of the property setter so the simulator can call it. What about making this a method on `Parameter`, maybe `def check_value(value)`? Then `Parameter.__init__` could call `self.check_value(value)` and the simulator code could call `param.check_value(value)`.

The exception message should maybe include the parameter name and the value, to help in locating the actual problem when the problem arises in the simulator calls (where the parameter name and value are often coming from a separate data source and so the call stack won't help identify the problem). Or you can leave these messages alone, then catch these exceptions in the simulator `_check_parameter_value` helper and re-raise them with augmented messages. Making these a custom subclass of ValueError would help with that.

FFroehlich: Good point, initially thought about adding to `__new__` which didn't work as assumptions were not set. Checking in `__ini__` does indeed work with a class method. Also realized that I already added testcases that checked for bad value in `__init__`, which already passed. `__init__` calls the setter so no additional checks are necessary there ;)

pysb/core.py
```diff
@@ -1268,20 +1268,32 @@ class Parameter(Component, Symbol):
...
         The provided value is converted to a float before being stored, so any
         value that cannot be coerced to a float will trigger an exception.
+    nonnegative : bool, optional
+        Sets the assumption whether this parameter is nonnegative (>=0).
+        Affects simplifications of expressions that involve this parameter.
+        By default, parameters are assumed to be non-negative.
+    integer : bool, optional
+        Sets the assumption whether this parameter takes integer values,
+        which affects simplifications of expressions that involve this
+        parameter. By default, parameters are assumed to take integer values.
```
alubbock: Isn't the default that parameters are **not** assumed to be integers?

FFroehlich: true, good catch!

PR #477, opened 2019-12-21
Faster ComplexPattern matching through canonical species representation

FFroehlich:


FFroehlich:
This PR implements a more efficient way of performing exact pattern matching on `ComplexPattern` using canonicalization. Canonicalization is implemented by sorting the respective `MonomerPatterns` by string (ignoring bond numbers) and then roundtripping through graph generation once to canonicalize bond numbers

PR #481, opened 2020-01-22
Use empty arrays when there are no observables or expressions

ortega2247:
When saving a simulation with include_obs_exprs=True and the model doesn't have expressions, PySB would save ```self._yexpr``` which is a list with np.ndarray with shape ```(len(self.tout[n]),)```. When we try to load this simulation the SimulationResult reshapes ```self._yexpr``` with ```len(tout[n]) * len(expr_names)``` but because expr_names is empty this would fail. This PR should fix that.

Steps to reproduce error:

```python

import numpy as np
from pysb.simulator import SimulationResult, ScipyOdeSimulator

tspan = np.linspace(0,100,100)
sim = ScipyOdeSimulator(model, tspan).run()
sim.save('tyson_sim.h5', include_obs_exprs=True)

sim2 = SimulationResult.load('tyson_sim.h5')

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: python
 sudo: required
 dist: trusty
 env:
-- PYVER=2.7
 - PYVER=3.6
 - PYVER=3.7
+- PYVER=3.8
 before_install:
 - wget http://repo.continuum.io/miniconda/Miniconda${PYVER:0:1}-latest-Linux-x86_64.sh
     -O miniconda.sh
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by alubbock** [comment]
Please add a unit test along with your PR. Here's one based on your example:

```python
import numpy as np
from pysb.simulator import SimulationResult, ScipyOdeSimulator
import io
from nose.tools import assert_raises

buff = io.BytesIO()
tspan = np.linspace(0,100,100)
sim = ScipyOdeSimulator(model, tspan).run()
sim.save(buff, include_obs_exprs=False)

sim2 = SimulationResult.load(buff)
assert len(sim2.observables) == len(tspan)
assert_raises(ValueError, lambda: sim2.expressions)
```

Thanks.

Issue #435, opened 2019-04-24
Drop theano support in ScipyOdeSimulator


alubbock:
Theano support in ScipyOdeSimulator will be dropped in PySB 2.0. Users should switch to the `compiler='cython'` backend instead.

jmuhlich:
Resolved by #482

Issue #434, opened 2019-04-24
Drop weave support in ScipyOdeSimulator


alubbock:
`weave` is only supported by Python 2.x, and we're dropping Python 2.7 support for PySB 2.0. Hence, `weave` support will be dropped too.

PR #482, opened 2020-01-22
Remove weave and theano from ScipyOdeSimulator

alubbock:
Remove legacy weave and theano support from ScipyOdeSimulator.
The Python exporter has also been updated to use Cython
rather than weave.


jmuhlich:
Closes #434, closes #435

PR #483, opened 2020-01-23
ScipyOdeSimulator speedup through use of partial

ortega2247:
Currently, PySB uses the concurrent.futures submit function for multiprocessing. This submit function adds the simulation function to a queue that is then accessed by a process worker. When we try to simulate large numbers (~10 000) of initial conditions or parameter values, the simulation takes a long time to finish. After digging into the python multiprocessing code, I think that the problem arises from having too many elements in the queue, which makes it slower for the workers to get() _CallItems and put() the simulation results in the result_queue queue. Using map would fix this as it divides the parameter values or initial conditions into different chunks that are then submitted to a queue, thus reducing the number of elements in a queue (although they have a larger size).

**Some benchmarks:**
Code:
```python
import numpy as np
from pysb.simulator import ScipyOdeSimulator

pars = np.array([p.value for p in model.parameters])
pars_100 = np.tile(pars, (100, 1))
pars_1000 = np.tile(pars, (1000, 1))
pars_10000 = np.tile(pars, (10000, 1))
tspan = np.linspace(0, 500, 500)
sim = ScipyOdeSimulator(model, tspan)


def simulate_tyson(datatype):

    if datatype == 'small':
        sim.run(param_values=pars_100, num_processors=4)

    elif datatype == 'medium':
        sim.run(param_values=pars_1000, num_processors=4)

    else:
        sim.run(param_values=pars_10000, num_processors=4)


if __name__ == '__main__':
    from timeit import Timer

    t1 = Timer("simulate_tyson('small')", "from __main__ import simulate_tyson")
    t2 = Timer("simulate_tyson('medium')", "from __main__ import simulate_tyson")
    t3 = Timer("simulate_tyson('big')", "from __main__ import simulate_tyson")

    print('Simulating with 100 pars            : ', t1.timeit(1))
    print('Simulating with 1000                : ', t2.timeit(1))
    print('Simulating with 10000               : ', t3.timeit(1))

```
**Results:**
PySB with concurrent futures submit, 20 cores, Linux

Simulating with 100 pars            :  1.553913926705718
Simulating with 1000                  :  12.606221742928028
Simulating with 10000               :  252.70697646960616

PySB with concurrent futures map, 20 cores, Linux

Simulating with 100 pars            :  1.4315921925008297
Simulating with 1000                  :  12.493971290066838
Simulating with 10000               :  121.1136589422822

**Benchmarking queues of different sizes:**
Code:
```python
import multiprocessing
import numpy as np

pars = np.array([p.value for p in model.parameters])
pars_100 = np.tile(pars, (100, 1))
pars_1000 = np.tile(pars, (1000, 1))


def with_queue(datatype):
    q = multiprocessing.Queue(1000)
    data = pars
    if datatype == 'small':
        for d in range(1000):
            q.put(data)
            q.get()

    else:
        for d in range(100000):
            q.put(data)
            q.get()


if __name__ == '__main__':
    from timeit import Timer

    t1 = Timer("with_queue('small')", "from __main__ import with_queue")
    t2 = Timer("with_queue('big')", "from __main__ import with_queue")

    print('Using queue.Queue with small data            : ', t1.timeit(1))
    print('Using queue.Queue with huge data             : ', t2.timeit(1))

```
**Results:**
Using queue.Queue with small data            :  0.08456832298543304
Using queue.Queue with huge data             :  8.2658322310308

### DIFF
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -21,6 +21,7 @@ import distutils
 import pysb.bng
 import sympy
 import re
+from functools import partial
 import numpy as np
 import warnings
 import os
@@ -518,25 +519,15 @@ class ScipyOdeSimulator(Simulator):
         else:
             self._logger.debug('Multi-processor (parallel) mode using {} '
                                'processes'.format(num_processors))
-        with SerialExecutor() if num_processors == 1 else \
-                ProcessPoolExecutor(max_workers=num_processors) as executor:
-            for n in range(n_sims):
-                results.append(executor.submit(
-                    _integrator_process,
-                    self._code_eqs,
-                    self._jac_eqs,
-                    num_species,
-                    num_odes,
-                    self.initials[n],
-                    self.tspan,
-                    self.param_values[n],
-                    self._init_kwargs.get('integrator', 'vode'),
-                    compiler=self._compiler,
-                    integrator_opts=self.opts,
-                    compiler_directives=self._compiler_directives
-                ))
-            trajectories = [r.result() for r in results]
 
+        with ProcessPoolExecutor(max_workers=num_processors) as executor:
+            sim_partial = partial(_integrator_process, code_eqs=self._code_eqs, jac_eqs=self._jac_eqs,
+                                  num_species=num_species, num_odes=num_odes, tspan=self.tspan,
+                                  integrator_name=self._init_kwargs.get('integrator', 'vode'),
+                                  compiler=self._compiler, integrator_opts=self.opts,
+                                  compiler_directives=self._compiler_directives)
+
+            trajectories = list(executor.map(sim_partial, *[self.initials, self.param_values]))
         tout = np.array([self.tspan] * n_sims)
         self._logger.info('All simulation(s) complete')
         return SimulationResult(self, tout, trajectories)
@@ -636,9 +627,8 @@ def _get_rhs(compiler, code_eqs, ydot=None, jac=None, compiler_directives=None):
     return rhs
 
 
-def _integrator_process(code_eqs, jac_eqs, num_species, num_odes, initials,
-                        tspan, param_values, integrator_name, compiler,
-                        integrator_opts, compiler_directives):
+def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species, num_odes, tspan,
+                        integrator_name, compiler, integrator_opts, compiler_directives):
     """ Single integrator process, for parallel execution """
     if compiler == 'python':
         code_eqs = sympy.lambdify(*code_eqs)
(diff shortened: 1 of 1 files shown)


pysb/simulator/scipyode.py
```diff
@@ -636,9 +627,8 @@ def rhs(t, y, p):
     return rhs
 
 
-def _integrator_process(code_eqs, jac_eqs, num_species, num_odes, initials,
-                        tspan, param_values, integrator_name, compiler,
-                        integrator_opts, compiler_directives):
+def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species, num_odes, tspan,
+                        integrator_name, compiler, integrator_opts, compiler_directives):
```
alubbock: Is there a reason to change the argument order here, since you're calling this function using keyword arguments anyway?

pysb/simulator/scipyode.py
```diff
@@ -689,17 +679,3 @@ def _integrator_process(code_eqs, jac_eqs, num_species, num_odes, initials,
         trajectory[i:, :] = 'nan'
 
     return trajectory
```
alubbock: I see your PR removed the `SerialExecutor`, so I was curious whether restoring it for the `num_processors=1` case would make things faster (lack of extra process overhead). On my Mac laptop it seemed to make little to no difference in practice, so I think we can safely remove it.

pysb/simulator/scipyode.py
ortega2247: Yes. executor.map is calling the function using *args  ```fs = [self.submit(fn, *args) for args in zip(*iterables)]``` So we need to put the param_values and initials at the beginning of the _integrator_process function

pysb/simulator/scipyode.py
ortega2247: We could keep the SerialExecutor and use the map function with only one chunk instead of the submit function. That would get rid of some overhead from the process submit.

pysb/simulator/scipyode.py
alubbock: Got it. Since this is a private function and there's a rationale for the order change, I think it's fine.

pysb/simulator/scipyode.py
alubbock: I think the overhead is negligible since only one process is launched. Happy to leave as is.

jmuhlich: I always assumed the speedup from SerialExecutor was mostly from avoiding pickling all the inputs and outputs for each task. Was the overhead measured with sufficiently large input and/or output?

jmuhlich:
This is odd, because `map` just calls `submit` under the hood anyway. There must be an issue with how we're using `submit` that is causing the slowdown...

In my own code, I choose `submit` over `map` when I want to do asynchronous progress reporting while the tasks are executing. That might actually be something we'd want to add here, so it might be worth trying to get the `submit` approach to run as fast as this new `map` approach.

One last thing - I just remembered why we should keep the SerialExecutor: Debugging with the asynchronous executors is a nightmare! With the SerialExecutor you can set breakpoints and step in/out of the simulator without any fuss.

alubbock:
@jmuhlich makes a good point about debugging - that seems like a good reason by itself to retain `SerialExecutor`.

It seems that it's @ortega2247's use of `partial` that's responsible for the speedup, rather than `map` per se. I've pushed a commit restoring the `SerialExecutor`, and used `submit` rather than `map` (the code basically mirrors what `map` does, but leaves the option of async progress reporting, as @jmuhlich suggests).

It also adds a `try`/`finally` block which cancels pending tasks, for better Ctrl+C support (which was supported by `map` anyway, but not in the current `master` branch).

What do you think, @jmuhlich and @ortega2247 ?

jmuhlich:
Good catch on the use of `partial` @alubbock! I wonder why that speeds things up...

Issue #414, opened 2019-02-05
End of Python 2.7 support


alubbock:
Python 2.x support will end on 01/01/20. Many packages are [pledging](https://python3statement.org/) to remove Python 2 support by then, or even sooner.

I propose we aim to remove Python 2 support by that date, which will also remove `weave` support. I'd propose also removing Theano support, since that library is now end of life.

jmuhlich:
Thumbs-up or -down this one to informally vote on this one. I'm all for it.

jmuhlich:
We should probably jump to PySB 2.0 as part of the schism.

alubbock:
As part of the Python version support policy decided in #463, it's been agreed that PySB will drop Python 2.x support on January 1, 2020.

PR #485, opened 2020-01-24
Import ABC from collections.abc instead of collections for Python 3.9 compatibility.

### LINKED ISSUE #484: Importing ABC from collections directly was removed in Python 3.9
https://bugzilla.redhat.com/show_bug.cgi?id=1794342

Sample instance but other instances also need to be fixed. I will raise a PR for this.

https://github.com/pysb/pysb/blob/54ff9e40ecb4835eae73287f1eb983e05b9cc960/pysb/importers/json.py#L40

tirkarthi:
Fixes #484

### DIFF
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -14,6 +14,7 @@ import abc
 from warnings import warn
 import shutil
 import collections
+from collections.abc import Sequence
 import pysb.pathfinder as pf
 import tokenize
 from pysb.logging import get_logger, EXTENDED_DEBUG
@@ -33,10 +34,6 @@ try:
 except NameError:
     basestring = str
 
-try:
-    from collections.abc import Sequence
-except ImportError:
-    from collections import Sequence
 
 def set_bng_path(dir):
     """ Deprecated. Use pysb.pathfinder.set_path() instead. """
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -33,6 +33,10 @@ try:
 except NameError:
     basestring = str
 
+try:
+    from collections.abc import Sequence
+except ImportError:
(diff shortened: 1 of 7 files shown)


pysb/bng.py
```diff
@@ -33,6 +33,10 @@
 except NameError:
     basestring = str
 
+try:
+    from collections.abc import Sequence
+except ImportError:
```
alubbock: We've [dropped Python 2.7 support](#414) recently, so there's no need for the `try`/`except` here. Simply importing from `collections.abc` should suffice. Applies to the other instances as well.

jmuhlich:
This has been deprecated since Python 3.3 anyway, so I'm all for fixing it.

pysb/bng.py
tirkarthi: Thanks, removed them. The `setup.py` still has the classifier Python 2 and hence I added these try/except block.

alubbock: Thanks. We're still in the process of removing Python 2 remnants from our codebase - thanks for bringing that one to my attention!

pysb/simulator/cupsoda.py
```diff
@@ -23,6 +23,10 @@
     import pycuda.driver as cuda
 except ImportError:
     cuda = None
+try:
+    from collections.abc import Iterable
+except ImportError:
+    from collections import Iterable
```
alubbock: One more instance to remove the try/except block from. Thanks!

tirkarthi: Thanks for the catch, removed.

alubbock:
Thanks again, looks good!

Issue #443, opened 2019-04-25
Proposal: increase precision when writing out parameters


alubbock:
When models are exported to BioNetGen or Kappa, parameters are printed using `%e`, which can lose precision:

    >>> '%e' % math.pi
    3.141593e+00

We could increase precision by using the `%.17g` format specifier:

    >>> '%.17g' % math.pi
    '3.1415926535897931'

Granted, `pi` is irrational, but clearly there are cases where this increased precision is desirable. As an example, it is causing trajectory disrepancies in round-tripped simulations using the BNG importer, e.g. for the `fceri_ji`, `test_paramname`, and `tlmr` models.

I propose we increase precision on the Kappa and BNG generators to `%.17g` as part of the move to PySB 2.0, but add an option to use the existing precision settings to maintain reproducibility with existing simulations. Perhaps as a kwarg `reduced_precision=True` on the relevant Simulators?

Initial discussion on this issue was on PR #410, where we decided to create track this issue separately.

PR #490, opened 2020-01-30
Double precision for BNG and Kappa params

alubbock:
As per #443, parameter values are not currently exported with full
precision to BNG and Kappa. This PR introduces 17 decimal digits,
in order to preserve full precision when round-tripping 64-bit
floats (as per IEEE 754 standard).

Since this might make previous simulations not exactly reproducible,
we might want to introduce a way of enabling users to use
reduced precision, either on a per-simulator or global level.
Open to suggestions on this. Might be complicated in the BNG case
by reaction network caching - local function parameters
read from BNG would need to be regenerated in the case
of a precision change.

Closes: #443


jmuhlich:
I found that the same Docker container (thus identical system libraries), same PySB version, same model, same parameter sets, on **different machines** (one Linux, one Windows) gave slightly different results already, so I wouldn't be too worried about reproducibility here.

pysb/tests/test_importers.py
```diff
@@ -16,10 +16,7 @@
 REDUCED_PRECISION = {
     'CaOscillate_Func': 1e-4,
     'michment': 1e-8,
-    'motor': 1e-8,
```
jmuhlich: Nice that this fixes most of the precision issues here!

alubbock:
> I found that the same Docker container (thus identical system libraries), same PySB version, same model, same parameter sets, on **different machines** (one Linux, one Windows) gave slightly different results already, so I wouldn't be too worried about reproducibility here.

Strange. Same Docker engine version? Did you manage to isolate the cause?

I'll merge this in, but an explanation of the change should be included in the release notes.

jmuhlich:
Didn't isolate the cause, but I know there are a few pieces of relevant FPU state that might not have been explicitly set by the code in the container or that can "leak" over from other processes outside the container during context switching. See more here:

https://randomascii.wordpress.com/2013/07/16/floating-point-determinism/

The differences were very small, but we were running a simulation to steady-state which is a case where tiny differences become readily noticeable. Even though we were using a tolerance for the convergence rather than waiting for all observables to be exactly equal for two steps, the two machines required an order of magnitude difference in simulated time (tspan) to settle to within that tolerance.

PR #491, opened 2020-01-31
Faster SimulationResult loading: use slices

ortega2247:
According to [h5py](https://h5py.readthedocs.io/en/stable/high/dataset.html#reading-writing-data):  Slice specifications are translated directly to HDF5 “hyperslab” selections, and are a fast and efficient way to access data in the file.

PySB currently passes the `trajectories` h5py Dataset object to the `SimulationResult` object. Then, it iterates over the trajectories. This iteration is slow because it uses the Dataset` __iter__` method. If we access the data first (which it already returns a numpy.array) and iterate over these objects, loading the simulations become faster.

Some benchmarks:

```python
from pysb.simulator import SimulationResult, ScipyOdeSimulator
import numpy as np

tspan = np.linspace(0, 100, 100)
pars = np.array([p.value for p in model.parameters])
all_pars_10000 = np.tile(pars, (10000, 1))
simulation = ScipyOdeSimulator(model, tspan).run(param_values=all_pars_10000, num_processors=4)
simulation.save('tyson_sims.h5')


def load_function():
    sim = SimulationResult.load('tyson_sims.h5')

if __name__ == '__main__':
    from timeit import Timer
    t1 = Timer("load_function()", "from __main__ import load_function")
    print('Loading sim with 10000 pars            : ', t1.timeit(1))
```
**Results on a Mac**

Slicing Datasets:
Loading sim with 10000 pars            :  1.49567398900035

Current PySB approach:
Loading sim with 10000 pars            :  53.9273291490008

### DIFF
--- a/pysb/simulator/base.py
+++ b/pysb/simulator/base.py
@@ -1237,10 +1237,10 @@ class SimulationResult(object):
             obs_and_exprs = None
 
             if 'observables' in dset.keys():
-                obs_and_exprs = list(dset['observables'])
+                obs_and_exprs = list(dset['observables'][:])
 
             if 'expressions' in dset.keys():
-                exprs = dset['expressions']
+                exprs = dset['expressions'][:]
                 if obs_and_exprs is None:
                     obs_and_exprs = list(exprs)
                 else:
@@ -1275,7 +1275,7 @@ class SimulationResult(object):
                 model=model,
                 initials=initials,
                 param_values=dset['param_values'][:],
-                tout=np.array(dset['tout']),
+                tout=dset['tout'][:],
                 trajectories=trajectories,
                 observables_and_expressions=obs_and_exprs,
                 squeeze=dset.attrs['squeeze'],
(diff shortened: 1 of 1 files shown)


ortega2247:
I just realize that we could slice expressions, observables and tout. This would speedup the loading time when the simulations are saved with observables and expressions

ortega2247:
Also, is there a reason to return the observables and expressions as lists instead of numpy arrays? There is a for loop in lines 1247-1251 of base.py in the simulator folder that slows down the loading time of large simulations with observables and expressions. If we could return a numpy array this for loop could be vectorized and that would also speed up the load function

jmuhlich:
@ortega2247 Do you want to add those improvements in this PR or start a new one?

ortega2247:
I am going to start a new PR, so this one can be merged

jmuhlich:
@ortega2247 could you please rebase this or merge master into it so GitHub shows it as up-to-date and mergeable? Thanks. In the future if you leave the "Allow edits from maintainers" box checked when creating a PR then pysb maintainers who aren't in LoLab-VU can do that for you.

ortega2247:
Sounds good! Thanks

PR #494, opened 2020-03-18
Implement BNG boolean treatment within expressions

alubbock:
BNG treats booleans as integers within expressions (1=True,
0=False), so an expression like `(2 > 1) * 4` will get evaluated
to `4`. Sometimes these expressions get produced by BNG
local functions and may be present in other imported models as
well. However, doesn't implement multiplication of
BooleanAtoms, which causes a TypeError.

This PR implements the BNG style behaviour on import by
patching BooleanTrue.__mul__ and BooleanFalse.__mul__.
Not very elegant, but it works.

### DIFF
--- a/pysb/bng.py
+++ b/pysb/bng.py
@@ -18,7 +18,7 @@ from collections.abc import Sequence
 import pysb.pathfinder as pf
 import tokenize
 from pysb.logging import get_logger, EXTENDED_DEBUG
-
+from sympy.logic.boolalg import BooleanTrue, BooleanFalse
 try:
     from cStringIO import StringIO
 except ImportError:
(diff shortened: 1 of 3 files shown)


jmuhlich:
I'd prefer a more robust solution rather than monkey-patching sympy internals.

Do you see literal constant `2 > 1`, or is it really `x > 1` which is evaluated later? Both can be fixed with a call to `replace` but we would need one extra step to fix the literal version. If we need to handle the literal one, we can add `evaluate=False` to the `parse_expr` call to avoid the "BooleanAtom not allowed in this context" exception at that point. If it's really only symbolic stuff like `x > 1` getting parsed then we can skip that step.

Once we have a sympy expression, we can replace all boolean expressions used in `Mul` operations as below. If other operators need this treatment we can refactor this a bit to make that easier to implement.

```python
# Helper to make the replace function below easier to read
def is_bool_expr(e):
    return isinstance(e, sympy.boolalg.Boolean) and not isinstance(e, sympy.AtomicExpr)

def fix_boolean_multiplication(e):
    args = [
        sympy.Piecewise((1, a), (0, True)) if is_bool_expr(a) else a
        for a in e.args
    ]
    return sympy.Mul(*args)

expr = expr.replace(sympy.Mul, fix_boolean_multiplication)
```

alubbock:
The expression that was causing the issue was of the form `some_expr*(0>0)` (with multiple different literal values in variants of the local expression). I got your fix to work with a slight modification:

```python
def fix_boolean_multiplication(*args):
    new_args = [
        sympy.Piecewise((1, a), (0, True)) if is_bool_expr(a) else a
        for a in args
    ]
    return sympy.Mul(*new_args)
```
This issue seems less likely to occur with Sympy 1.6, where one couldn't write a local expression in the above form in the first place. It's unclear to me whether an expanded local expression read back from BNG could be in that form when the original is not, but it might be possible.

I've pushed the fix as a new commit. Thanks!

PR #500, opened 2020-05-09
Add timeout argument to the ScipyOdeSimulation run function

ortega2247:
In the Lopez lab we have found out that during model calibration, the fitting algorithm can sample parameters that generate stiff dynamics that take too long to integrate. Thus this PR takes advantage of the concurrent futures implementation to set a timeout for each simulation submitted to the ProcessPoolExecutor

### DIFF
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -18,7 +18,7 @@ import logging
 import itertools
 import contextlib
 import importlib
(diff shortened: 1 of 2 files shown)


ortega2247:
Running Futures don't stop after the timeout, it only raises a TimeoutError but the Future keeps running until it completes so it doesn't work as we were expecting

PR #502, opened 2020-05-13
Fix observable trajectories on HDF5 load

alubbock:
When loading a SimulationResult from an HDF5 file using the .load()
method, observables are left as zeroes (if the observables were not
saved with the results using the include_obs_exprs=True argument).

The issue is due to the fact that Observable species and coefficients
are not saved in the JSON representation, despite the
include_netgen=True argument. This was missed due to a typo in the
SimulationResult unit tests, also fixed here.

This PR addresses the issue by calculating the Observable species
and coefficients using PySB's built-in pattern matching during
the JSON import process. This approach allows existing HDF5 files
to be used without resaving.

### DIFF
--- a/pysb/export/json.py
+++ b/pysb/export/json.py
@@ -233,6 +233,13 @@ class PySBJSONWithNetworkEncoder(PySBJSONEncoder):
         return rxn
 
     @classmethod
+    def encode_observable(cls, obs):
+        o = super(PySBJSONWithNetworkEncoder, cls).encode_observable(obs)
+        o['species'] = obs.species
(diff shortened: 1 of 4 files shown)


jmuhlich:
Shouldn't the JSON export code *be* saving observable species and coefficients when include_netgen=True? Or do we always want to regenerate that information on load? Any accidental divergence in our pattern matcher vs. the BNG observable generation logic will lead to hard-to-diagnose issues when reloading these JSON dumps.

jmuhlich:
Cool. Will approve after CI passes.

PR #507, opened 2020-06-04
fix typehints PEP0484

FFroehlich:
see https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values


jmuhlich (approved):
Seems OK. Does this help given that all the downstream places these `kwargs` get passed are still only hinted as `dict`?

FFroehlich:
> Seems OK. Does this help given that all the downstream places these `kwargs` get passed are still only hinted as `dict`?

This fixes the syntax highlighting in PyCharm when specifying Parameters. Downstream usage doesn't matter but could also be fixed, could  be helpful for developers.

PR #513, opened 2020-08-07
fix sbml export with unset compartment size

FFroehlich:


pysb/export/sbml.py
```diff
@@ -141,7 +141,10 @@ def convert(self, level=(3, 2)):
                 _check(c)
                 _check(c.setId(cpt.name))
                 _check(c.setSpatialDimensions(cpt.dimension))
-                _check(c.setSize(cpt.size.value))
+                if cpt.size is None:
```
jmuhlich, line 144: Suggest this one-liner instead, what do you think?
```python
_check(c.setSize(1 if cpt.size is None else cpt.size.value))
```

FFroehlich: Yes, definitely more pythonic!

jmuhlich (approved):

PR #518, opened 2020-09-24
Extend function support in BNGL import and pysb export

FFroehlich:


pysb/importers/bngl.py
```diff
@@ -42,7 +70,7 @@ def __init__(self, filename, force=False, cleanup=True):
 
         self.model.name = os.path.splitext(os.path.basename(filename))[0]
         self._force = force
-        self._model_env = {}
+        self._model_env = BNGL_FUNCTIONS
```
alubbock, line 73: Is enumerating the functions like this necessary? It seems to work on current `master` branch already. Here's a test program:

```python
from pysb.importers.bngl import model_from_bngl
import tempfile

bngl_test = """
begin seed species
  A 100
  B 100
end seed species

begin reaction rules
  A() + B() <-> C() 0.01, 1
end reaction rules

# not used in model, only for import tests
begin functions
  f1() = exp(10)
  f2() = atanh(10)
  f3() = sqrt(log2(50))
end functions
"""

with tempfile.NamedTemporaryFile('w', suffix='.bngl', encoding='utf8') as tf:
    tf.write(bngl_test)
    tf.flush()
    model = model_from_bngl(tf.name)

print(model)

for expr in model.expressions:
    print(f'{expr.name}: {expr.expr.evalf()}')
```

FFroehlich: It's been a while since I wrote that code, so I am not sure. I am surprised that log2 actually works, since this definitely used to require the mapping. This might be due to recent changes by @jmuhlich or sympy added `log10` and `log2` functions function.

Issue #524, opened 2020-12-10
Compartment size as Expression?


dweindl:
PySB does not allow to specify compartment sizes as Expressions:

https://github.com/pysb/pysb/blob/c434f6ab98301beee1bf9d2a5093f0c79da78824/pysb/core.py#L1375-L1376

Is there a specific reason why only Parameters are allowed? It would be convenient if also (constant) Expressions could be provided.

jmuhlich:
There's no good reason! When I originally wrote that code we didn't have Expression support, and we never revisited it once Expressions were introduced. BNG can take a constant expression there, so we can easily fix this.

dweindl:
Thanks @jmuhlich , I'll submit a PR shortly.

PR #525, opened 2020-12-10
Allow constant expressions for compartment sizes

dweindl:
Previously, only parameters were allowed for compartment sizes.
This change allows specifying compartment sizes using constant expressions.

Closes pysb/pysb#524

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1345,9 +1345,10 @@ class Compartment(Component):
     dimension : integer, optional
         The number of spatial dimensions in the compartment, either 2 (i.e. a
         membrane) or 3 (a volume).
-    size : Parameter, optional
-        A parameter object whose value defines the volume or area of the
-        compartment. If not specified, the size will be fixed at 1.0.
+    size : Parameter or Expression, optional
+        A parameter or constant expression object whose value defines the
+        volume or area of the compartment. If not specified, the size will be
+        fixed at 1.0.
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


jmuhlich:
This is great, thanks! Would you mind adding a new test to exercise the new code path (and might as well check the old one too since we have no test for this). You would edit `pysb/tests/test_core.py` and add something like this (not actually tested!)

```python
@with_model
def test_compartment():
    Parameter('A', 1.0)
    Parameter('B', 2.0)
    Expression('E', A * B)
    Compartment("C1")
    Compartment("C2", C1, 2, A)
    Compartment("C3", C1, 2, E)
```

jmuhlich:
Oh, also `test_invalid_compartment` should get an additional check to assert that a dynamic expression is not valid:

```python
Monomer('A')
Observable('O', A)
Expression('E', O)
assert_raises(Exception, Compartment, 'c2', size=E)
```

dweindl:
Will do, thanks for the pointers!

dweindl:
Tests added as suggested. Worked without changes. The updated `test_compartment` fails with pysb from master (as intended) and passes here. `test_invalid_compartment` passes on master as well as here.

jmuhlich:
Looks like we have some CI failures due to unrelated Sympy changes. I'll work out how to fix those and re-run CI for this PR.

jmuhlich (approved):

jmuhlich:
Thanks Daniel!

dweindl:
Thank you!

PR #527, opened 2020-12-17
Split simulator egfr benchmark into codegen and run

jmuhlich:
This lets us track codegen and simulation performance separately. Also:

* Switch from setup to setup_cache to reduce setup overhead.
* Remove Earm10 test with empty cython_directives since it's broken
  after a previous refactoring and is about to be obviated by an upcoming
  improvement in ScipyOdeSimulator.


jmuhlich:
Sorry, I just realized the egfr_extended codegen and simulation benchmarks that this PR introduces are completely new. I created this branch by cherry-picking some commits from another branch where I'd introduced a combined egfr_extended simulator benchmark, and it was so long ago that I forgot that benchmark was only on that branch!

jmuhlich:
I edited the PR title to make more sense in light of this revelation.

alubbock (commented):
Ran into a couple of issues getting `asv run` to work with these changes; see inline comments

PR #528, opened 2020-12-31
Fix BNG code generation for Eq and Ne

jmuhlich:
Previously, Eq generated `Eq(...)` instead of `==`, and similarly for Ne.
This changes them to generate `==` and `!=`, as BNGL (muparser) expects.
The code is copied and simplified from StrPrinter's _print_Relational
implementation.

### DIFF
--- a/pysb/generator/bng.py
+++ b/pysb/generator/bng.py
@@ -4,6 +4,7 @@ import pysb
 from pysb.core import MultiState
 import sympy
 from sympy.printing import StrPrinter
+from sympy.printing.precedence import precedence
 
 # Alias basestring under Python 3 for forwards compatibility
 try:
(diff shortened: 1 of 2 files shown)


pysb/generator/bng.py
```diff
@@ -308,6 +309,18 @@ def _print_And(self, expr):
     def _print_Or(self, expr):
         return super(BngPrinter, self)._print_Or(expr).replace('|', '||')
 
+    def _print_Relational(self, expr):
+        if getattr(expr, "rel_op", None) not in {"==", "!=", "<", "<=", ">", ">="}:
+            raise NotImplementedError(
```
alubbock, line 314: Does this exclude anything other than in-place assignments? Could add tests for this Exception, if there's anything relevant.

alubbock (approved):

pysb/generator/bng.py
jmuhlich: I was more guarding against new relational operators being added to sympy in the future. I accessed `rel_op` defensively since it feels a little like an internal implementation detail even though it's not underscore-prefixed, and I was afraid it might change in the future.

I actually think those `Assignment` classes mentioned in  `StrPrinter._print_relational` (referenced by `str` names, not actual `class` objects mind you) don't even exist anymore, or at least not as subclasses of `Relational`. You can find classes with those names under `sympy.codegen.ast` but they don't extend `Relational` and they don't have a `rel_op`.

jmuhlich: I mean, I could certainly write a test that defines a new `Relational` subclass and ensure that printing it raises, if you think that's useful.

jmuhlich: Done!

PR #531, opened 2021-01-12
Scipyode simulator refactor

jmuhlich:
This moves all codegen out of the simulator class. A new abstract class
RhsBuilder is introduced to manage codegen, and Python and Cython concrete
implementations are provided. The codegen itself is now performed using sympy
expression manipulation and sympy's own codegen rather than hand-rolled string
manipulations and Cython.inline. Performance is improved since Cython.inline's
overhead on long code strings was really hurting us.

### DIFF
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -419,30 +419,23 @@ class RhsBuilder:
         self.p = sympy.MatrixSymbol('p', len(model.parameters_all()), 1)
         self.e = sympy.MatrixSymbol('e', len(expr_constant), 1)
         self.o = sympy.MatrixSymbol('o', len(model.observables), 1)
-        species_subs = {
-            sympy.Symbol('__s%d' % i): self.y[i]
-            for i, _ in enumerate(model.species)
-        }
+        # Parameters symbols. We also need this independently of all_subs.
         param_subs = dict(zip(model.parameters_all(), self.p))
-        obs_subs = dict(zip(model.observables, self.o))
-        expr_dynamic_subs = {
-            e: e.expand_expr()
-            for i, e in enumerate(expr_dynamic)
-        }
-        expr_constant_subs = dict(zip(expr_constant, self.e))
-        obs_matrix = scipy.sparse.lil_matrix(
-            (len(model.observables), len(model.species)), dtype=np.int64
-        )
-        for i, obs in enumerate(model.observables):
-            obs_matrix[i, obs.species] = obs.coefficients
+        # All substitution rules we need to apply to the reaction expressions.
+        all_subs = param_subs.copy()
+        # Species symbols.
+        all_subs.update({
+            sympy.Symbol('__s%d' % i): self.y[i]
+            for i in range(len(model.species))
+        })
+        # Observables symbols.
+        all_subs.update(dict(zip(model.observables, self.o)))
+        # Constant expressions symbols.
+        all_subs.update(dict(zip(expr_constant, self.e)))
+        # Dynamic expressions (expanded).
+        all_subs.update({e: e.expand_expr() for e in expr_dynamic})
         self.kinetics = sympy.Matrix([
-            r['rate']
-            .xreplace(expr_constant_subs)
-            .xreplace(expr_dynamic_subs)
-            .xreplace(param_subs)
-            .xreplace(species_subs)
-            .xreplace(obs_subs)
-            for r in model.reactions
+            r['rate'].subs(all_subs) for r in model.reactions
         ])
         if with_jacobian:
             # The Jacobian can be quite large but it's extremely sparse. We can
@@ -457,6 +450,11 @@ class RhsBuilder:
             self._logger.debug("Computing Jacobian matrices")
             self.kinetics_jacobian_y = kinetics_sparse.jacobian(self.y)
             self.kinetics_jacobian_o = kinetics_sparse.jacobian(self.o)
+        obs_matrix = scipy.sparse.lil_matrix(
+            (len(model.observables), len(model.species)), dtype=np.int64
+        )
+        for i, obs in enumerate(model.observables):
+            obs_matrix[i, obs.species] = obs.coefficients
         self.observables_matrix = obs_matrix.tocsr()
         self.stoichiometry_matrix = model.stoichiometry_matrix
         self.model_name = model.name
--- a/pysb/simulator/scipyode.py
+++ b/pysb/simulator/scipyode.py
@@ -204,8 +204,6 @@ class ScipyOdeSimulator(Simulator):
                                            _run_kwargs=[])
         n_sims = len(self.param_values)
 
-        num_species = len(self._model.species)
-        num_odes = len(self._model.odes)
         if num_processors == 1:
             self._logger.debug('Single processor (serial) mode')
         else:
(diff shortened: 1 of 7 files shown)


pysb/simulator/scipyode.py
```diff
@@ -551,6 +417,353 @@ def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species,
...
+        self.cleanup = cleanup
+        expr_dynamic = model.expressions_dynamic(include_derived=True)
+        expr_constant = model.expressions_constant(include_derived=True)
+        self.y = sympy.MatrixSymbol('y', len(model.species), 1)
+        self.p = sympy.MatrixSymbol('p', len(model.parameters_all()), 1)
+        self.e = sympy.MatrixSymbol('e', len(expr_constant), 1)
+        self.o = sympy.MatrixSymbol('o', len(model.observables), 1)
+        species_subs = {
+            sympy.Symbol('__s%d' % i): self.y[i]
+            for i, _ in enumerate(model.species)
```
alubbock, line 490: Minor quibble, but I prefer `for i in range(len(model.species))`

pysb/simulator/scipyode.py
```diff
@@ -551,6 +417,353 @@ def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species,
...
+        }
+        param_subs = dict(zip(model.parameters_all(), self.p))
+        obs_subs = dict(zip(model.observables, self.o))
+        expr_dynamic_subs = {
+            e: e.expand_expr()
+            for i, e in enumerate(expr_dynamic)
```
alubbock, line 496: No need for `enumerate` here, just `for e in expr_dynamic`

pysb/simulator/scipyode.py
```diff
@@ -551,6 +417,353 @@ def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species,
...
+        )
+        for i, obs in enumerate(model.observables):
+            obs_matrix[i, obs.species] = obs.coefficients
+        self.kinetics = sympy.Matrix([
+            r['rate']
+            .xreplace(expr_constant_subs)
+            .xreplace(expr_dynamic_subs)
+            .xreplace(param_subs)
+            .xreplace(species_subs)
+            .xreplace(obs_subs)
```
alubbock, line 510: Would it be more efficient here to store the various subs as a generator (e.g. using `itertools.chain`) to allow a single `xreplace` call? This would obviate the need for multiple loops over the symbols in `r['rate']`, but maybe it wouldn't be a significant speedup. Just a thought.

pysb/simulator/scipyode.py
```diff
@@ -551,6 +417,353 @@ def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species,
...
+            code_gen,
+            filepath=self.work_path,
+            extra_compile_args=[
+                # The RHS evaluation is such a tiny part of overall integration
+                # time, even for huge models, that compiler optimization
+                # actually takes more time than it will ever yield back. Since
+                # Cython sets -O2 by default we need to override it.
+                "-O0",
+                # TODO: Enable this line once Cython 3.0 is released.
+                #"-DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION",
```
alubbock, line 675: Should this be handled with runtime Cython version detection?

pysb/simulator/scipyode.py
```diff
@@ -551,6 +417,353 @@ def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species,
...
+            ],
+        )
+        escaped_name = re.sub(
+            r"[^A-Za-z0-9]",
+            "_",
+            self.model_name.encode("unicode_escape").decode(),
+        )
```
alubbock, line 682: I guess there might be some edge cases where this causes a mapping collision, but since a unique temp directory is used this shouldn't cause any issues.

pysb/simulator/scipyode.py
```diff
@@ -551,6 +417,353 @@ def _integrator_process(initials, param_values, code_eqs, jac_eqs, num_species,
...
+        state["_jacobian_fn"] = None
+        return state
+
+    @property
+    def work_path(self):
+        """Location for saving any temporary working files."""
+        if self._work_path is None:
+            # Only initialize this on demand since it will create a directory
+            # which requires cleanup on destruction.
+            self._work_path = tempfile.mkdtemp(suffix="_pysb_compile")
```
alubbock, line 554: Might be good to log this path at `debug` level

alubbock (changes requested):
Looks good, thanks for this. Only a couple of minor things to address and check.

pysb/simulator/scipyode.py
```diff
@@ -286,11 +191,13 @@ def _patch_distutils_logging(self):
     def _test_cython(cls):
         if not hasattr(cls, '_use_cython'):
             cls._use_cython = False
-            if Cython is None:
-                return
             try:
-                Cython.inline('x = 1', force=True, quiet=True)
+                import Cython
+                Cython.inline("x = 1", force=True, quiet=True)
+                del Cython
```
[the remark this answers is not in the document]
jmuhlich: It's just to keep it our module namespace clean.

pysb/simulator/scipyode.py
jmuhlich: I was just going to make us depend on `Cython>=3` once it's out. Thoughts on that?

jmuhlich: I guess I could make it opportunistic right now, I'll give it a shot. No good way to test it though.

pysb/simulator/scipyode.py
jmuhlich: The performance isn't an issue but I'll rework it to only call `xreplace` once per reaction anyway. `xreplace` takes a dict so `chain` wouldn't work, but I can just merge all the dicts and use that. I need to preserve entry order though -- should I use an OrderedDict or rely on the official dict ordering in python 3.7+? What python releases will we want to support?

jmuhlich:
All sensible requests. Will work on these ASAP.

pysb/simulator/scipyode.py
jmuhlich: I might sprinkle a few more logging calls around, now that you mention it.

pysb/simulator/scipyode.py
jmuhlich: Oops, `xreplace` doesn't handle the case where the result of one substitution needs to also be processed by a later one. `subs` does handle this, but I thought the performance was worse. I'll benchmark before committing to an approach.

pysb/simulator/scipyode.py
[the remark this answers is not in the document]
jmuhlich: I ended up going back to the old way, both for better error logging if Cython isn't installed and to support the Cython 3.0+ check down below.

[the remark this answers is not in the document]
jmuhlich: I ended up factoring out the whole compiler detection/verification code. Will push soon after I check the behavior in various environments (this is hard to test automatically as I'm sure you know).

pysb/simulator/scipyode.py
```diff
@@ -483,8 +483,18 @@ def __getstate__(self):
         state["_expressions_constant_fn"] = None
         state["_rhs_fn"] = None
         state["_jacobian_fn"] = None
+        # Loggers are not picklable in Python 3.6.
+        state["_logger_extra"] = state["_logger"].extra
```
jmuhlich, line 487: While it would be fine to just store this back in `_logger`, I felt it was nice to be explicit that it's not storing the actual logger.

alubbock (approved):

PR #539, opened 2021-04-14
Use python dict to support bngl hash type arguments in console actions

ortega2247:
This adds support to use bngl hash type arguments like max_stoich that are passed to generate_network. I use f-strings because i feel it is easier to read, but most of the code in pysb uses the old string formatting (% operator). Comments, feedback, and requests are more than welcome

```python 
from pysb.examples.bax_pore import model
from pysb.bng import generate_equations

generate_equations(model, **{'max_stoich':{"BAX":1}})
model.species
```
That caps the number of BAX molecules in a complex to 1


alubbock (changes requested):
Looks good - please could you add a small test for this to the BNG test suite? Similar to the example you give on the PR. Thanks!

ortega2247:
Just added the requested test. Also, I change the exporters so that  the `generate_equation` is only called  when the model has not generated the equations before. I think that can be good for large models that take a long time to generate equations and so we don't have to generate equations again if we want to export the model to some other format. This also facilitates the export of models' equations that has been generated using arguments like max_stoich

pysb/export/json.py
```diff
@@ -243,7 +243,8 @@ def encode_model(cls, model):
         d = super(PySBJSONWithNetworkEncoder, cls).encode_model(model)
 
         # Ensure network generation has taken place
-        generate_equations(model)
+        if not model.odes:
```
alubbock, line 246: `generate_equations` already internally checks if the rxn network has been generated, so these `if` statements aren't required

pysb/bng.py
```diff
@@ -137,8 +137,18 @@ def _format_action_args(cls, **kwargs):
             Arguments and values to supply to BNG
         """
         if kwargs:
-            action_args = ','.join('%s=>%s' % (k, BngConsole._bng_param(v))
-                                   for k, v in kwargs.items())
+            processed_pars = []
+            for k, v in kwargs.items():
+                bng_param = BngConsole._bng_param(v)
+                if isinstance(bng_param, dict):
```
alubbock, line 143: Ideally, this should be `if isinstance(bng_param, collections.Mapping)` since it supports e.g. `OrderedDict`

pysb/bng.py
```diff
@@ -137,8 +137,18 @@ def _format_action_args(cls, **kwargs):
...
+                    par_string = f'{k}' + '=>{' + ','.join(f"{k}=>{v}" for k, v in bng_param.items()) + '}'
+                    processed_pars.append(par_string)
```
alubbock, line 145: This line could be factored out - it applies to both `if/else` branches

pysb/tests/test_bng.py
```diff
@@ -23,6 +23,19 @@ def test_generate_network():
...
 
+@with_model
+def test_generate_network_max_agg():
+    Monomer('A', ['b1', 'b1'])
+    Parameter('k', 1)
+    Parameter('A_0', 1)
+    Initial(A(b1=MultiState(None, None)), A_0)
+    Rule('a_bind_a', A(b1=MultiState(None, WILD)) + A(b1=MultiState(None, WILD)) >>
+         A(b1=MultiState(1, WILD)) % A(b1=MultiState(1, WILD)), k)
+    generate_equations(model, **{"max_stoich": {"A": 2}})
```
alubbock, line 34: Why not simply `generate_equations(model, max_stoich={"A": 2})`?

alubbock (changes requested):
Thanks @ortega2247. I've made a few small suggestions in the inline comments.

pysb/export/json.py
ortega2247: that's right

pysb/bng.py
ortega2247: Sounds good. Thanks for the feedback

pysb/bng.py
ortega2247: yep

pysb/tests/test_bng.py
ortega2247: that's right

pysb/bng.py
```diff
@@ -137,8 +137,16 @@ def _format_action_args(cls, **kwargs):
             Arguments and values to supply to BNG
         """
         if kwargs:
-            action_args = ','.join('%s=>%s' % (k, BngConsole._bng_param(v))
-                                   for k, v in kwargs.items())
+            processed_pars = []
+            for k, v in kwargs.items():
+                bng_param = BngConsole._bng_param(v)
+                if isinstance(bng_param, collections.Mapping):
+                    par_string = f'{k}' + '=>{' + ','.join(f"{k}=>{v}" for k, v in bng_param.items()) + '}'
```
jmuhlich, line 144: This is unwieldy and not really using the benefit of f-strings. Also re-using the local `v` variable is a bit confusing. How about something like this (not tested) :
```python
args_string = ",".join(f"{k}=>{v}" for k, v in bng_param.items())
par_string = f"{k}=>{args_string}"
```

jmuhlich, line 144: Should we perform some validation on the `kwargs` keys and values here to make sure we don't emit invalid BNGL? If the BNG parser fails due to invalid BNGL here the error will probably be hard for the user to make sense of. The keys should be validated against the regex that matches whatever BNG accepts there, but I'm not sure what the value syntax is.

jmuhlich (changes requested):

PR #540, opened 2021-04-14
render_species with MultiState

ortega2247:
Add support to MultiState in the render_species function


alubbock (commented):
Thanks for implementing. Two thoughts:

* Is there a way to visually indicate a `MultiState`, e.g. with a different colour?
* I realise there aren't tests yet for the various render functions, but it'd be great to start adding some for new code.

PR #542, opened 2021-05-04
Fix Anaconda download link in docs

alubbock:


jmuhlich:
Should we link to https://www.anaconda.com/pricing instead which has options for both commercial and individual?

jmuhlich (approved):

PR #553, opened 2021-11-04
Energy modeling

jmuhlich:
This adds support for BNG's energy modeling formalism. Models can now contain EnergyPatterns as well as energy-based Rules with the Arrhenius rate law parameterized by activation energy (E₀) and rate distribution (φ). Energy-based models can be simulated directly in PySB or exported to BNGL for use with BioNetGen-based tools. The PySB JSON export protocol number is increased to 2 due to the addition of these new model features.


jmuhlich:
This probably still needs more tests but I wanted to get this out there to get the review process started.

alubbock (approved):

PR #559, opened 2021-12-14
Add total_rate option to rules for NFsim compatibility

rasi:
Hi, I had sent a pull request for this functionality a few years back (https://github.com/pysb/pysb/pull/304). I have now addressed the reviewer comment (https://github.com/pysb/pysb/pull/304#issuecomment-336250126). Let me know if you would like any further edits for this to be merged with master. Thanks!

### DIFF
--- a/pysb/core.py
+++ b/pysb/core.py
@@ -1432,6 +1432,13 @@ class Rule(Component):
         co-transport anything connected to that Monomer by a path in the same
         compartment. If False (default), connected Monomers will remain where
         they were.
+    total_rate: bool, optional
+        If True, the rate is considered to be macroscopic and is not
+        multiplied by the number of reactant molecules during simulation.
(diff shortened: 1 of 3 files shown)


alubbock (approved):
Thanks @rasi. Although your tests use `BngFileInterface`, I'd recommend using `pysb.simulator.BngSimulator` for most use cases, since it's a more full featured, higher level interface (e.g. it can loop through sets of initials/parameters). But happy to get this merged in.

jmuhlich:
The JSON export/import still needs to be updated to include this property -- this will require incrementing the protocol number. I already needed to increment it to 2 for energy support, so I might just submit a quick PR to add total_rate under protocol 2 before we make an official release. Also shouldn't we be able to parse total_rate in the BNG importer as well?

jmuhlich:
Also the `Rule.__repr__` needs to add `total_rate`.

PR #560, opened 2022-02-02
Update tutorial.rst

mustafaozen:
Updated dissociation rule "tBid_from_C8Bid". Corrected typo: Replaced "%" by "+" on the right-hand side of the rule as this is an example of dissociation.


jmuhlich (approved):
Thanks for fixing this.

PR #570, opened 2022-12-16
Enable pattern matching with compartments

FFroehlich:
Enable `RulePatternMatcher` and `SpeciesPatternMatcher` for models with compartments. Support was already there, just removed `NotImplementedError` - this is validated by including` move_connected`, which contains species with compartments, in the test_all_species generated test.


alubbock (commented):
LGTM overall, though it's been a while since I worked with compartments. Ideally it'd be good to see test coverage for the two lines highlighted by codecov - particularly pysb/core.py line 956.

FFroehlich:
> LGTM overall, though it's been a while since I worked with compartments. Ideally it'd be good to see test coverage for the two lines highlighted by codecov - particularly pysb/core.py line 956.

Ah! turns out adding the edges in `956` is actually not even necessary, simply adding the species compartment node to the graph is sufficient. Removed the respective lines, added a test for the value error and simplified the code a bit further.

FFroehlich:
@alubbock @lh64

alubbock (approved):

PR #574, opened 2023-03-28
Fix numpy v1.24 compatibility

alubbock:
Numpy v1.24 deprecated ragged arrays unless dtype=object is passed (numpy/numpy#22004). The SimulationResult class needs to support different length entries in tout in the case these differ across simulations. PySB shouldn't convert these to a numpy array, instead preferring to keep them as a list/iterable, which avoids causing an error with Numpy v1.24 when running the PySB test suite.

### DIFF
--- a/pysb/simulator/base.py
+++ b/pysb/simulator/base.py
@@ -748,7 +748,7 @@ class SimulationResult(object):
             self.run_kwargs = {}
 
         self.squeeze = squeeze
-        self.tout = np.asarray(tout)
+        self.tout = tout
         self._yfull = None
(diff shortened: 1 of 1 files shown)


JamesPino (commented):
Seems like an easy fix. At first I was curious what prevented tout and trajectories from being lists of different sizes, but found the checks later on. I don't see the check for observables_and_expressions. Would/could that ever occur? It's independent of this fix, but was just looking at how self.tout is used later and noticed.

jmuhlich (approved):

PR #581, opened 2023-07-17
Dependency pins for sympy and sphinx

jmuhlich:
Pin sympy<1.12 and sphinx<7 to avoid breaking changes.

### DIFF
--- a/.github/pysb-conda-env.yml
+++ b/.github/pysb-conda-env.yml
@@ -8,7 +8,7 @@ dependencies:
   - alubbock::kappa
   - numpy
   - scipy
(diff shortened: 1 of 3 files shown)


alubbock (approved):
Looks like it's struggling to solve the conda environment on Ubuntu/Python 3.7, but 3.7 is end of life anyway. From my perspective, we can go ahead and merge this and remove Python 3.7 from tests in a future PR.

jmuhlich:
We might want to consider using mamba or micromamba to manage the test environments, since its solver is extremely fast compared to conda.

PR #584, opened 2024-01-23
update py versions in CI matrix

### LINKED ISSUE #582: remove python 3.7 and add 3.11 to CI matrix

FFroehlich:
fixes #582


jmuhlich:
`nose` is unmaintained and breaks under python 3.11. Allegedly someone forked it to `pynose` which works, so I'll try that.

jmuhlich (approved):
Looks good.

PR #587, opened 2024-02-23
Disable GitHub Actions fail-fast

jmuhlich:
This lets us see all the possible problems across the entire CI matrix rather than cancelling all the jobs as soon as any one fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ jobs:
     name: Tests & Coverage
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


jmuhlich:
I have no idea why Ubuntu / Python3.9 fails at `pip install libroadrunner`. I tested that same combination locally and it works fine. You can observe that the wheel does exist on pypi.

FFroehlich (approved):

FFroehlich:
> I have no idea why Ubuntu / Python3.9 fails at `pip install libroadrunner`. I tested that same combination locally and it works fine. You can observe that the wheel does exist on pypi.

reran tests, looks like the libroadrunner failure was transient.

PR #588, opened 2024-02-27
Switch MacOS CI runner to ARM

jmuhlich:
Mostly straightforward changes to the matrix `os` entry to get the guaranteed arm64 runner with `macos-14`. I think since I started this work, `macos-latest` is now arm64, but sticking with `macos-14` for now seems fine.

I needed to rework how stochkit presence/absence is handled since it no longer works on macOS at all (no arm64 stochkit build). My solution sets an environment variable in the steps that install stochkit, then checks that variable in the actual nosetests step to conditionally exclude stochkit tests. I needed to use the nosetests `--exclude` option because that's the only way to also skip specific doctests (there's no way to apply the `-a` attrs to doctests).

I also updated some action versions while I was editing the workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        os: ["ubuntu-latest", "macos-14", "windows-latest"]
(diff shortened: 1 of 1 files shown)


jmuhlich:
Need to resolve #596 to get this to pass CI.

FFroehlich (approved):

PR #595, opened 2024-06-18
Adapt to numpy 2.0

bgyori:
Numpy 2.0 was just released and I'm going over some repos to see if there are any incompatibilities. I found a couple of minor changes needed here to be compatible with the new version per https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide.


jmuhlich (approved):
Thanks, seems simple enough.

PR #599, opened 2024-09-02
Update sympy and numpy compatibility

jmuhlich:
This makes us compatible with sympy 1.12+ and numpy 2.0+.


pysb/bng.py
```diff
@@ -924,6 +924,47 @@ def _is_bool_expr(e):
     return isinstance(e, Boolean) and not isinstance(e, sympy.AtomicExpr)
 
 
+# BNG's expression syntax supports multiplication with a boolean, e.g.
+# `2 * (x > 0)`, but sympy does not. Our overall strategy is to convert
```
FFroehlich, line 928: Wouldn't there be the same issue for other operands such as `/` or `pow`?

pysb/bng.py
```diff
@@ -924,6 +924,47 @@ def _is_bool_expr(e):
...
+    return result
+
+
+# Helper function for _convert_mul.
+def _transform_mul(tokens, local_dict, global_dict):
+    # Strongly based on sympy.parsing.sympy_parser._transform_equals_sign.
+    result = []
+    if (tokenize.OP, '*') in tokens:
+        local_dict['__bngl_mul'] = sympy.Function('bngl_mul')
+        result.append((tokenize.NAME, '__bngl_mul'))
```
FFroehlich, line 954: Won't this break operator precedence as everything following will be executed after `bngl_mul`? I think it might be easier to directly transform `<` and `>` (and '=<' and '=>' if they are allowed in BNG) here as operator precedence wouldn't be an issue, it would't require any post-parsing processing and it would give compatibility other operands.

pysb/bng.py
```diff
@@ -924,6 +924,47 @@ def _is_bool_expr(e):
...
+# function call. Then, after parsing and evaluation, we perform a subexpression
+# replacement on the 'bngl_mul' calls that converts Boolean args to a Piecewise
+# and emits a normal Mul. This can't be done entirely in a parser transformation
+# because sympy hasn't yet parsed the expression to determine the AST structure
+# that lets us identify boolean subexpressions, and it can't be done entirely
+# post-parsing because multiplication of booleans raises an immediate error
+# during the evaluation step in parse_expr. This function is the parser
+# transformation.
+def _convert_mul(tokens, local_dict, global_dict):
+    res1 = sympy_parser._group_parentheses(_convert_mul)(tokens, local_dict, global_dict)
```
FFroehlich, line 941: this recursively applies this function to contents of parentheses groups, similar to traversing the AST, right?

FFroehlich (changes requested):

pysb/bng.py
jmuhlich: Yeah, but our previous code here only ever fixed up `*` so I could consider this change just a fix for the new stricter sympy semantics. Adding support for the other operators should be a separate change. In practice I think the `*` version of this is the only one that's been seen "in the wild" in BNG models so far.

pysb/bng.py
jmuhlich: I assume you mean converting all relational expressions to the corresponding 0 or 1 using a `Piecewise`? I don't think this works because we need to support BNG `if` expressions which we also convert to a sympy `Piecewise`, and I believe the conditions in a Piecewise needs to be booleans. I'll try it though.

pysb/bng.py
jmuhlich: Yes. Sympy already provides this implementation to handle similar use cases like converting `a = b` to `Eq(a, b)`.

pysb/bng.py
jmuhlich: You're absolutely right on the operator precedence problem with the current implementation though. Not sure what I was thinking there. I'll add a test case to catch this.

jmuhlich: > I believe the conditions in a Piecewise needs to be booleans

I think I solved this by making the `bngl_if` implementation explicitly treat both `False` and `0` as false (and anything else as true):
```python
expr = expr.replace(
    sympy.Function('bngl_if'),
    lambda cond, t, f: sympy.Piecewise(
        (f, sympy.Eq(cond, False) | sympy.Eq(cond, 0)),
        (t, True),
    )
)
```

pysb/bng.py
jmuhlich: I just realized that `convert_equals_signs` transform is actually pretty sketchy and doesn't even work properly when the `=` expression is in certain contexts, so it was kind of a bad example to learn from.

PR #600, opened 2024-10-18
Support PyBioNetGen provided BNGPATH

FFroehlich:
pypi: https://pypi.org/project/bionetgen/ github: https://github.com/RuleWorld/PyBioNetGen

could also help simplify BNG interface in the future


FFroehlich:
the python package does not provide the `Validate` directory with example models, so we have to stick with the conda installation in tests for now

jmuhlich (approved):

PR #601, opened 2024-10-18
Support for explicit time dependency in expressions

### LINKED ISSUE #534: Add support for explicit time in expressions
We will define a new `SpecialSymbol` subclass of `sympy.Symbol` that just serves to manage special/implicit variables like "time".  In pysb.core we will define an instance of `SpecialSymbol` named "time" and models will need to utilize this in any expression to refer to time. Exporters, importers, and simulators will need to be adjusted to support time.

FFroehlich:
fixes #534

### DIFF
--- a/pysb/tests/test_integrate.py
+++ b/pysb/tests/test_integrate.py
@@ -109,14 +109,14 @@ def test_integrate_with_expression():
     Rule('R2', None >> s20(), ks0)
     Rule('R3', s16() + s20() >> s16() + s1(), keff)
 
-    tspan = np.linspace(0, 40)
+    time = np.linspace(0, 40)
 
-    solver = Solver(model, tspan)
+    solver = Solver(model, time)
     solver.run()
 
-    assert solver.yexpr_view.shape == (len(tspan),
+    assert solver.yexpr_view.shape == (len(time),
                                        len(model.expressions_dynamic()))
-    assert solver.yobs_view.shape == (len(tspan), len(model.observables))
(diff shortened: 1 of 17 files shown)


pysb/tests/test_integrate.py
```diff
@@ -109,14 +111,29 @@ def test_integrate_with_expression():
...
     solver.run()
 
-    assert solver.yexpr_view.shape == (len(time),
+    assert solver.yexpr_view.shape == (len(tspan),
                                        len(model.expressions_dynamic()))
-    assert solver.yobs_view.shape == (len(time), len(model.observables))
+    assert solver.yobs_view.shape == (len(tspan), len(model.observables))
+
+@with_model
+def test_integrate_with_time():
```
jmuhlich, line 124: The `Solver` class that this file tests should have been deprecated in favor of the newer and better-designed Simulator classes, and in fact Solver just delegates to `ScipyOdeSimulator` now anyway. Could you move this test to `test_simulator_scipy.py`, and replace the Solver call with the following (including a check on the simulation result):
```python
    sim = ScipyOdeSimulator(model, tspan=tspan)
    simres = sim.run()
    assert np.isclose(simres.species[-1, 0], <expected final concentration of A to 6 figures>)
```

jmuhlich (changes requested):
Sorry I took so long to review this! Overall this looks clean and thorough, just a few minor suggestions.

jmuhlich (approved):
Looks great, thanks again!

PR #614, opened 2026-03-29
fix(ci): fix CI perf., Windows test discovery

### LINKED ISSUE #597: CI for Windows doesn't actually run any tests
The CI log for Windows runners shows the following final nosetests output:
```Ran 0 tests in 0.000s```
Somehow it's not finding any tests to run. I can't tell how long this has been happening since GHA doesn't seem to keep old logs around for very long.

alubbock:
GH Actions was running extremely slowly on Linux and Mac, and not discovering tests at all on Windows.

Fixes in this PR to get CI working and performant:

* Fix Windows CI running 0 tests by switching from `nosetests .` to `nosetests pysb --traverse-namespace`.

* Skip the test_multiprocessing_lambdify() test for now. This test was either not completing or extremely slow. SciPy v1.17 introduced a bug in handling analytic jacobians: scipy/scipy#24933. I opened a PR to fix this scipy/scipy#24934 which has now been merged, but not yet released.

* Guard run_* example scripts with `__main__` check to prevent side effects when importing during test discovery, which was also causing slowdowns.

* Skip SBML import tests when atomizer (sbmlTranslator) binary is unavailable, and remove atomizer from Windows conda env as it has no upstream support and the Windows build on alubbock channel is very old.

Minor housekeeping also included:

* GitHub Actions version bumps:
  - macos-14 runner to macos-latest (in line with other OSes)
  - checkout@v4 action to v6
  - codecov-action@v4 action to v6

* Remove the Anaconda defaults channels from CI (use conda-forge instead).

* Remove extraneous print() calls in tests.

---
Fixes #597

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,20 +9,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "macos-14", "windows-latest"]
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
         python-version: ["3.11", "3.12", "3.13", "3.14"]
     defaults:
       run:
         shell: bash -l {0}
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - uses: conda-incubator/setup-miniconda@v3
         with:
           miniconda-version: "latest"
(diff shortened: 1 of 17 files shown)


FFroehlich (approved):
👍

PR #617, opened 2026-03-31
feat: native network generator for PySB

alubbock:
Add pysb.netgen module, a pure-Python network generator implementation that expands rules to explicit species and reactions network without external dependencies.

Core implementation (pysb/netgen.py):
- NetworkGenerator class: iterative network expansion from seed species
- Canonical species hashing via Morgan-like label refinement and branch-and-bound permutation search w/ O(1) identity lookup
- Incremental rule firing: only apply rules to newly discovered species
- Raises NotImplementedError for energy rules and Tags (local fxns).

Model integration (pysb/core.py):
- Model(auto_netgen=True): lazy network generation on first access of model.species / model.reactions / model.reactions_bidirectional
- _DirtyList and _DirtyComponentSet wrappers track mutations to model.initials and model.rules, regenerating the network on demand

Simulator integration (pysb/simulator/base.py):
- New Simulator(netgen=...) parameter: 'auto' (default), 'bng', 'pysb'
- 'auto' uses pre-existing equations if present, 'pysb' if `auto_netgen=True` is set on the model, 'bng' otherwise.

Documentation (doc/modules/netgen.rst):
- Quickstart examples, feature comparison table vs pysb.bng, performance benchmarks, limitations, auto_netgen usage, BNG validation helpers
- Generated here on readthedocs for review: https://pysb--617.org.readthedocs.build/en/617/modules/netgen.html

Test suite (pysb/tests/test_netgen.py, plus a couple in core and for the Simulator base class):
- 20 PySB built-in examples validated against BNG output
- 34 BNG Validate directory models validated against BNG output
- Unit tests for private helpers and edge cases

Performance (MacBook Air M4, BioNetGen 2.9.1):
- Faster than BNG for all tested models. Smaller models do comparatively better due to subprocess and i/o overhead of calling BNG.
- Benchmarking script included in benchmarks/netgen_benchmark.py for independent verification.


FFroehlich:
This is great, especially that you figured out how to do good species hashing, that was always a pain point for me when I tried implementing this. Will have a look, but will likely need a few days to give feedback.

